Issue #4, opened 2016-05-31
integer to bigint


kwent:
We had already two occurrences when we hit the maximum value for some primary key who was `integer` type (value is `2147483647`). Cf. https://www.postgresql.org/docs/9.1/static/datatype-numeric.html

Would be nice to warn the user and let them create table with primary key `bigint` type right away cause it's really painful to change the column type later on ...

PR #5, opened 2016-05-31
Check if default primary key :integer type is used

kwent:
### Check if default primary key :integer type is used

- To address https://github.com/ankane/strong_migrations/issues/4

**Notes**:
 
- Might not be enough to just check if `:id` flag is false but should cover most of cases... We should check in the block, all column with `primary_key: true` flag and the `:limit` value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ gem 'strong_migrations'
 
 ## Dangerous Operations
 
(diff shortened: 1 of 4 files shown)


ankane:
Thanks @kwent.  For most situations, `id` as an integer rather than a bigint should be fine.

PR #6, opened 2016-06-04
add full safety for a file and add a test

### LINKED ISSUE #3: No way to disable for migration as a whole
Very excited about this gem and adding it to a large app. Unfortunately, I am stuck combing through hundreds of migrations to mark them as safe. Is there an easier way, such as marking the whole migration as safe without having to edit individual lines?

sgringwe:
Fixes #3


ankane:
Thanks for the PR @sgringwe. However, I'm not sure it's needed.  See my comment on #3 for an explanation.

PR #7, opened 2016-07-20
Update README.md

oliviertassinari:
Fix the `WARNING: there is no transaction in progress`.


ankane:
Thanks for the PR @oliviertassinari. `disable_ddl_transaction!` is an alternative way, but I personally find it's easier to miss in docs since it's outside the `change` method.

PR #8, opened 2016-08-04
adding remove_index to unsafe_migrations

aravind88:


ankane:
Hey @aravind88, thanks for the PR. 👍 

I don't think I would classify removing indexes as an unsafe operation (at least in practice, I haven't seen this bite anyone as I have the others).

PR #9, opened 2016-12-06
Avoid N+1 on add new default column

barnett:
Use `#pluck` over `#map`.

Note: I am unsure why `map` was used here, there might be some insight I missed.


ankane:
Hey @barnett, thanks for the PR 👍 Find in batches yields an array of records, so the existing method avoids an N+1. I believe pluck with throw an error, since array won't respond to it.

barnett:
Thanks for the response!

PR #13, opened 2017-05-17
Leverage Rails 5 ignored_columns

jturkel:
Update the "Removing a column" section to include details on the `ActiveRecord::Base.ignored_columns` method that landed in https://github.com/rails/rails/pull/21720.


ankane:
Great, thanks @jturkel 👍

PR #15, opened 2017-05-22
add_reference support

bartoszkopinski:


ankane:
Looks great, and complete with tests!! 💯  Thanks @bartoszkopinski

ankane:
It looks like `index: true` is the default value for Rails 5+. However, this won't raise an error if the `index` option is omitted.

ankane:
Fixed here: https://github.com/ankane/strong_migrations/commit/fb30970b0e9e6c9ec48deba47b48adb60dbbb36d

bartoszkopinski:
@ankane thanks 👍

PR #16, opened 2017-05-26
Add assume_safety_prior_to_version config option

### LINKED ISSUE #12: Easy way to ignore unsafe operations in pre-existing migrations
My codebase has 164 migrations already, many of which include "unsafe" operations (likely created before the app was even in production).

Is there an easy way to "assure safety" for all migrations before a certain version number?

vergenzt:
Fixes #12 and #3.

Let me know what you think!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - Added check for `force` option with `create_table`
 - Added `auto_analyze` option
+- Added `assume_safety_prior_to_version` option
 
 ## 0.1.6
 
--- a/lib/strong_migrations/migration.rb
+++ b/lib/strong_migrations/migration.rb
@@ -70,7 +70,11 @@ module StrongMigrations
     private
 
     def safety_overridden?
-      @safe || ENV["SAFETY_ASSURED"]
+      @safe || ENV["SAFETY_ASSURED"] || should_assume_safety_due_to_version?
+    end
+
+    def should_assume_safety_due_to_version?
+      version && version <= StrongMigrations.assume_safety_prior_to_version
     end
 
     def postgresql?
--- a/lib/strong_migrations/migration.rb
+++ b/lib/strong_migrations/migration.rb
@@ -14,7 +14,7 @@ module StrongMigrations
     end
 
     def method_missing(method, *args, &block)
-      unless @safe || ENV["SAFETY_ASSURED"] || is_a?(ActiveRecord::Schema) || @direction == :down
+      unless safety_overridden? || is_a?(ActiveRecord::Schema) || @direction == :down
         case method
         when :remove_column
           raise_error :remove_column
@@ -68,6 +68,10 @@ module StrongMigrations
     end
 
     private
+
+    def safety_overridden?
+      @safe || ENV["SAFETY_ASSURED"]
+    end
 
     def postgresql?
       %w(PostgreSQL PostGIS).include?(connection.adapter_name)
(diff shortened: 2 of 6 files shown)


vergenzt:
Adding changelog and readme entries right now, will amend.

ankane:
Hey @vergenzt, thanks for the PR 👍  It looks like it's causing the tests for fail. Also, left a few comments inline.

test/strong_migrations_test.rb
```diff
@@ -207,6 +221,18 @@ def test_create_table_force
...
+  def test_old_remove_column_assumed_safe
+    StrongMigrations.assume_safety_prior_to_version = SAFETY_ASSUMED_VERSION_CUTOFF
+    assert_safe OldRemoveColumnAssumedSafe
+    StrongMigrations.assume_safety_prior_to_version = 0
+  end
+
+  def test_new_remove_column_not_assumed_safe
+    StrongMigrations.assume_safety_prior_to_version = SAFETY_ASSUMED_VERSION_CUTOFF
+    assert_unsafe NewRemoveColumnNotAssumedSafe
+    StrongMigrations.assume_safety_prior_to_version = 0
```
ankane: Let's use `ensure` here.

CHANGELOG.md
```diff
@@ -32,3 +33,5 @@
 ## 0.1.0
 
 - First release
+
+[@vergenzt]: https://github.com/vergenzt
```
ankane: The current changelog doesn't list contributors. Let's keep it consistent.

vergenzt: Sorry about that. Last repo I contributed to did this and I didn't think about it.

vergenzt:
Thanks for the review @ankane! As I was looking at the tests I realized that success or failure of the tests depended on execution order, so I refactored the tests a bit in 2b005e6 to wipe the database and create the users table before each test case so nothing bleeds through. I believe I've addressed all comments.

Lmk what you think!

ankane:
Thanks for addressing. There should be a way to make tests work without wiping the DB (as they do now). I'd prefer to keep the changeset to a minimum.

vvondra:
This would be much appreciated!

ankane:
Hey @vergenzt, I cleaned this up and merged in 50f9827d5bc580dae5ec8c4c24bd822f283e0ee3. Thanks for driving this. I think it'll be a very useful feature. 👍

lib/strong_migrations/migration.rb
```diff
@@ -69,6 +69,14 @@ def method_missing(method, *args, &block)
 
     private
 
+    def safety_overridden?
+      @safe || ENV["SAFETY_ASSURED"] || should_assume_safety_due_to_version?
+    end
+
+    def should_assume_safety_due_to_version?
+      version && version <= StrongMigrations.assume_safety_prior_to_version
```
nisanth074: @vergenzt @ankane In what cases is `version` nil?

I just added the strong_migrations gem to our Rails 3 app. Unfortunately, the `start_after` option doesn't seem to work. Old dangerous migrations fail despite me configuring `start_after` option https://gist.github.com/nisanth074/ea45c8cdd50d4a08b53cb97b91fc3a6b

I added a `binding.pry` statement above this line and ran `bundle exec rake db:create db:migrate` again. It seems like `version` is always `nil` for all our migrations. So, `should_assume_safety_due_to_version?` returns nil and strong_migrations raises an error.

Can you let me know in what cases is `version` nil? Is Rails 3 supported?

Thanks for your time.

ijones16:
Thanks for your work on this feature. It helped me out with a 5 year old project 🙌

jpwhitaker:
I just found this thread after installing [rpush](https://github.com/rpush/rpush).

I had 28 migrations that pass the strong_migration check, and the rpush installation added 17 more that fail.  Is there any method like this that would allow me to skip a range of migrations?

Screen Shot 2021-01-18 at 11 45 02 AM

Issue #17, opened 2017-05-30
Consider running arbitrary SQL a dangerous operation


jturkel:
This gem provides awesome protection against dangerous operations when using the Rails migration DSL. Unfortunately it's easy for developers to inadvertently bypass this safety net when running arbitrary SQL e.g.

```
class AddNumWheelsToCars < ActiveRecord::Migration
  def up
    execute('ALTER TABLE cars ADD COLUMN num_wheels integer DEFAULT 4 NOT NULL')
  end

  def down
    remove_column(:cars, :num_wheels)
  end
end
```
It would be great if `execute` was considered a dangerous operation that required developers to explicitly indicate that they know what they're doing. I'd be happy to put up a PR if you're open to the change.

ankane:
Hey @jturkel, great idea. Happy to accept a PR. We do something similar for `change_table`.

PR #18, opened 2017-05-30
Add check for executing arbitrary SQL

jturkel:
This fixes https://github.com/ankane/strong_migrations/issues/17


ankane:
Beautiful, thanks @jturkel 💯

Issue #14, opened 2017-05-18
Rails 5 in_batches method


vlymar:
First off, thanks a ton for this fantastic guide.

I was recently following the instructions for [adding a column](https://github.com/ankane/strong_migrations#adding-a-column-with-a-default-value) and wondered if there was a cleaner way to batch update yet.

Turns out in rails 5 there's a new [in_batches](http://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-in_batches) method that can yield enumerators of relation objects, which lets you rewrite 

```ruby
User.find_in_batches do |users|
  User.where(id: users.map(&:id)).update_all some_column: "default_value"
end
```

as 

```ruby
User.in_batches.update_all some_column: "default_value"
```

If you also prefer the aesthetics of this new method, I can go ahead and open a PR with that option for Rails 5. But if you don't want to clutter the guide with more "For Rails < 5 use this, for Rails >= 5 use that", I understand. Just thought I'd share this new method :)

ankane:
Hey @vlymar, thanks for sharing. It's definitely has a cleaner look. I'm happy to accept a PR to replace the current suggestion.

PR #19, opened 2017-06-02
Add #in_batches code example for Rails 5+ batch processing.

vlymar:
PR for https://github.com/ankane/strong_migrations/issues/14 .

[#in_batches](http://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-in_batches) makes batch makes batch processing a bit cleaner.

Honestly not sure its worth cluttering up the snippet to add this, and since the method was only added in rails 5 I don't think it should replace the previous example.

Feel free to reject this PR. I'm pretty enthusiastic about #in_batches but that seems tangential to the concern of safe migrations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,10 @@ class AddSomeColumnToUsers < ActiveRecord::Migration
     # 2
     commit_db_transaction
 
-    # 3
+    # 3.a (Rails 5+)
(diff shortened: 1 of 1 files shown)


ankane:
Thanks @vlymar 👍 Rails 4.2 is no longer supported by the Rails team, so we can remove that path in the next few months.

PR #21, opened 2017-07-12
Fix race condition in algorithm for adding a column with a default

jturkel:
The algorithm for adding a column with a default will miss rows added after the backfill completes (step 3) and before the column default is set (step 4). This can be fixed by setting the column default before backfilling values.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,9 +39,9 @@ Also checks for best practices:
 ### Adding a column with a default value
 
 1. Add the column without a default value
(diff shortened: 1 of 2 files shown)


ankane:
Great find, thanks @jturkel 💯

jturkel:
Thanks @ankane. Any plans to do a release soon?

ankane:
No, but you can use `master` if you want to use it immediately.

PR #22, opened 2017-08-10
Mark remove index operation as dangerous

zentetsukenz:
Since removing index in Postgresql requires an exclusive lock on table. This operation can block migration and cause downtime on a highly usage table.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,3 +18,5 @@ matrix:
     - gemfile: test/gemfiles/mysql2.gemfile
       env: ADAPTER=mysql2
       rvm: 2.4.1
+addons:
+  postgresql: "9.2"
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ gem 'strong_migrations'
 - renaming a column
 - removing a column
 - executing arbitrary SQL
-- adding an index non-concurrently (Postgres only)
+- adding/removing an index non-concurrently (Postgres only)
 - adding a `json` column to an existing table (Postgres only)
 
 For more info, check out:
@@ -112,7 +112,7 @@ end
 
 Once it’s deployed, create a migration to remove the column.
 
-### Adding an index (Postgres)
+### Adding/Removing an index (Postgres)
 
 Add indexes concurrently.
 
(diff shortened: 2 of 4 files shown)


ankane:
Hey @zentetsukenz, thanks for the PR 👍  The code looks great. I've been aware of the option to drop concurrently, but haven't personally experienced issues without it. Have you run into anything?

Also, looks like there are some limitations:

> an index that supports a UNIQUE or PRIMARY KEY constraint cannot be dropped this way

https://www.postgresql.org/docs/current/static/sql-dropindex.html

zentetsukenz:
Hi @ankane 

Sorry for not mentioned the problem in the PR.

We had recently experienced a fail migration upon running the `remove_index` without concurrently on an intensively used table while deploying. Since dropping an index requires an exclusive lock on a table which when run on a table that have many sessions connect to it, PG will throw a deadlock error, `PG::TRDeadlockDetected` , since the lock is blocked by other sessions.

On a second thought, I think this can be just a warning since it won't cause a downtime but just that the `remove_index` migration can be stuck in the middle of a deployment.

For the limitation, not sure what we should do. Would love to have your suggestion about this.

ankane:
I believe a lock timeout will fix that. https://github.com/ankane/strong_migrations#lock-timeout-postgres

zentetsukenz:
👌 will check that out

zentetsukenz:
Close for now as it seems like table lock can be avoid by using Postgres lock timeout.

krainboltgreene:
Despite the insight on the lock timeout, I definitely think this gem should still provide a warning.

krainboltgreene:
Also, I think strong_migrations should also warn the user to provide a column property so they can safely rollback as well.

oehlschl:
Came across this issue today as well after attempting to remove an index non-concurrently for one of our most used tables (users), not realizing that it would wait to acquire a full table lock. Obviously we should know what we're doing, but I agree with @krainboltgreene that it still makes sense to warn users of unsafe operations even if the problem isn't guaranteed or could be solved with a lock_timeout.

"Better safe than sorry" seems like the overarching point of this library, and there are enough issues that removing indexes seems to still be painful.

PR #27, opened 2017-10-04
fix typos

ghiculescu:


ankane:
Thanks @ghiculescu 👍  Not sure the 2nd one is really a typo, but it works...

PR #31, opened 2017-11-01
Leverage disable_ddl_transaction!

jturkel:
This PR updates the recipe for adding indexes concurrently to leverage [disable_ddl_transaction!](http://api.rubyonrails.org/classes/ActiveRecord/Migration.html#method-c-disable_ddl_transaction-21) which is the Rails recommended way of disabling transactions in migrations. 

Note `disable_ddl_transaction!` was introduced in Rails 4.0 so this code won't work in Rails 3.2 but the `algorithm` option for `add_index` was also introduced in Rails 4.0 in this [commit](https://github.com/rails/rails/commit/e199dc1a570d4f0d9a07628268835bce5aab2732) so the recipe was already incorrect for Rails 3.2. Perhaps it's time to drop support Rails 3.2?

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,8 +119,9 @@ Add indexes concurrently.
 
 ```ruby
 class AddSomeIndexToUsers < ActiveRecord::Migration
+  disable_ddl_transaction!
+  
   def change
-    commit_db_transaction
     add_index :users, :some_index, algorithm: :concurrently
   end
 end
--- a/lib/strong_migrations/migration.rb
+++ b/lib/strong_migrations/migration.rb
@@ -148,9 +148,9 @@ Once that's deployed, wrap this step in a safety_assured { ... } block."
   end"
         when :add_index
 "Adding a non-concurrent index locks the table. Instead, use:
+  disable_ddl_transaction!
 
   def change
-    commit_db_transaction
     add_index :users, :some_column, algorithm: :concurrently
   end"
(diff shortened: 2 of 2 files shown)


ankane:
Hey @jturkel, thanks for the PR 👍  I'd prefer to keep it inline (with `commit_db_transaction`) as it's easier to understand what to do. `disable_ddl_transaction!` can be missed since it's outside the `change` method.

jturkel:
I've had a few developers who were confused because the suggestions from the gem aren't consistent with the Rails [Migration Guide](http://guides.rubyonrails.org/active_record_migrations.html) or [Migration API Doc](http://api.rubyonrails.org/classes/ActiveRecord/Migration.html) but I definitely see your point about keeping closer to the `change` method. Your call. Either way this is an awesome gem.

nate00:
@ankane what do you think about making the error messages configurable? The configurability could be as simple as:
- Extract the big case statement into a `def error_message(message_key)` method that users could override.
- Treat that method and the `message_key` symbols as a public interface (i.e., only make changes between major versions, and document the changes in the CHANGELOG).

And, of course, more sophisticated configuration interfaces are possible.

At my job we already override a few messages to make them consistent with our migration style guide.

I'm happy to submit a PR if that helps!

ankane:
As customizable error messages is the most requested feature, added on master.

PR #32, opened 2017-11-01
Make error message for removing columns consistent with README

jturkel:
This PR updates the error message for removing columns to be consistent with the README changes introduced in https://github.com/ankane/strong_migrations/pull/13.


ankane:
Awesome, thanks @jturkel 👍

PR #33, opened 2017-11-10
Add warning for remove_index

ianforsyth:
Removing an index also puts a lock on the table in postgresql. Unfortunately, rails doesn't allow us to pass `algorithm: concurrently` into the `remove_index` option so we can't be as robust as the `add_index` warning. The solution is to execute it raw.

Let me know your thoughts, love the gem!

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ The following operations can cause downtime or errors:
 - renaming a column
 - removing a column
 - adding an index non-concurrently (Postgres only)
+- removing an index non-concurrently (Postgres only)
 - adding a `json` column to an existing table (Postgres only)
 
 For more info, check out:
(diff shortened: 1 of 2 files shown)


ianforsyth:
If this is something you'd consider adding, I can update the specs too

ankane:
Hey @ianforsyth, thanks for the PR 👍  Check out #8 and #22.

ianforsyth:
@ankane Ah, okay, I'll close this one down, we haven't actually run into any trouble with it. Thanks for maintaining this project, it's really helpful!

ankane:
No problem, thanks for taking the time to make it better.

mkilling:
We’ve just been bitten by this when dropping indexes from a production PostgreSQL database:

```
PG::TRDeadlockDetected: ERROR:  deadlock detected
DETAIL:  Process 5691 waits for AccessExclusiveLock on relation 20512 of database 16402; blocked by process 5687.
Process 5687 waits for AccessShareLock on relation 16475 of database 16402; blocked by process 5691.
HINT:  See server log for query details.
```

We did drop the indexes without `algorithm: :concurrently` assuming that `strong_migrations` would warn us if we did anything dangerous. Now we know better, but I would love to revive this discussion and see if adding a warning for `remove_index` when run without `algorithm: :concurrently` or a `lock-timeout`.

ankane:
Hey @mkilling, recent versions warn if you don't have a lock timeout set. You can also opt-in to requiring indexes to be removed concurrently (see the readme for how to do that).

mkilling:
@ankane Oh, I somehow missed that. Very handy. Thanks for the quick reply!

Issue #36, opened 2018-02-10
[Proposal]: unsafe change_column_null default value


twonegatives:
Rails [change_column_null](https://apidock.com/rails/ActiveRecord/ConnectionAdapters/SchemaStatements/change_column_null) method accepts four arguments, with the fourth one replacing existing NULLs with some other value. Looks kinda useful feature although the implementation of such a replacement might be pretty dangerous on large tables ([postgres](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb#L436-L443), [mysql](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/abstract_mysql_adapter.rb#L358)):
```ruby
UPDATE #{quote_table_name(table_name)}
SET #{quote_column_name(column_name)}=#{quote_default_expression(default, column)}
WHERE #{quote_column_name(column_name)} IS NULL
```
What we're doing here is a `WHERE` query on a table which might be lacking of needed index on `column_name` and `UPDATE` for (potentially) millions of records at once.
A little more safe way might be to require user to make this `UPDATE` in batches manually before firing `change_column_null` up.
WDYT?

ankane:
Hey @twonegatives, makes sense to me. Do you want to try adding it?

twonegatives:
hey @ankane sure, I'll give it a try.

ankane:
Merged!!

PR #37, opened 2018-02-18
Added check for change_column_null

twonegatives:
related issue: https://github.com/ankane/strong_migrations/issues/36


ankane:
Awesome, big thanks @twonegatives 💯One minor tweak: since `default_value` is set with `to_s`, it will never be `nil`. I added a test case for this and addressed in 0cd68c558fc71e014f9b33ba782ac72d3224f32a.

twonegatives:
Nice catch :)

PR #40, opened 2018-03-05
Add support for Active Record 3.x and set ruby 2 as the lowest supported version

jcmfernandes:
Feedback is more than welcome.


jcmfernandes:
After some experimentation, I found out that this isn't enough. Whereas ActiveRecord >= 4 correctly passes the migration version to each migration object - https://github.com/rails/rails/blob/master/activerecord/lib/active_record/migration.rb#L987 -, ActiveRecord 3.x does not - https://github.com/rails/rails/blob/3-2-stable/activerecord/lib/active_record/migration.rb#L538

This means that `StrongMigrations::Migration#version_safe?` always returns falsy, regardless of what's in `StrongMigrations.start_after`. Monkey-patching `ActiveRecord::MigrationProxy` (when on ActiveRecord 3.x) seems like only way forward. What do you think?

ankane:
Hey @jcmfernandes, thanks for the PR 👍 Unfortunately, the gemspec is a bit outdated (purposefully for backward compatibility). I'm only actively supporting ActiveRecord 4.2+. Happy to add a min Ruby version of 2.2 (as others are EOL) and bump the min ActiveRecord when 0.3.0 is released.

PR #41, opened 2018-04-06
users.map(&:id) -> users.ids for README.md

vredniy:


ankane:
Hey @vredniy, thanks for the PR 👍 However, I don't think this works.

vredniy:
It should work, 'cause `ids` method has been added in Rails 4.0.2 https://apidock.com/rails/v4.0.2/ActiveRecord/Calculations/ids @ankane

PR #46, opened 2018-06-28
PostGISAdapter enhancements

chadwilken:
If `PostGISAdapter` is defined patch columns method with `StrongMigrations::AlphabetizeColumns` module.

PostGIS [includes](https://github.com/rgeo/activerecord-postgis-adapter/blob/master/lib/active_record/connection_adapters/postgis_adapter.rb#L39) their own schema statements which override the default columns implementation from Rails.


ankane:
Thanks @chadwilken 👍

PR #48, opened 2018-08-10
Change documentation on add_column_default

AlaRuba:
I was wondering if we could change the documentation of the error on `add_column_default` to a recommendation to backfill date on existing rows. `add_column_default` would have backfilled the data, so I think it's good to have an equivalent set of operations as a recommendation.

### DIFF
--- a/lib/strong_migrations.rb
+++ b/lib/strong_migrations.rb
@@ -18,11 +18,19 @@ module StrongMigrations
 the entire table to be rewritten.
 
 Instead, add the column without a default value,
-then change the default.
+then change the default, then make sure to backfill
+the data of existing rows.
 
   def up
     add_column :users, :some_column, :text
     change_column_default :users, :some_column, \"default_value\"
+    # Make sure to backfill the data
+    safety_assured do
+      execute %{
+        UPDATE users
+        SET some_column = \"default_value\"
+      }
+    end
   end
 
   def down
--- a/lib/strong_migrations.rb
+++ b/lib/strong_migrations.rb
@@ -19,23 +19,35 @@ the entire table to be rewritten.
 
 Instead, add the column without a default value,
 then change the default, then make sure to backfill
-the data of existing rows.
+the data of existing rows in another migration.
 
+# First migration
+class AddSomeColumnToUsers < ActiveRecord::Migration[5.2]
   def up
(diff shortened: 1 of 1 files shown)


ankane:
Hey @AlaRuba, thanks for the PR 👍 I think it makes sense to add a note that you need to backfill, but it needs to be a new migration. Check out the [existing docs](https://github.com/ankane/strong_migrations#adding-a-column-with-a-default-value) for an explanation.

AlaRuba:
Hey @ankane thanks for the feedback. I decided to just copy the example you have in your docs. I think repeating it in the error message makes it more obvious.

ankane:
Hey @AlaRuba, thanks for updating. I think we should customize the message for the ActiveRecord version they're using.

AlaRuba:
Is this what you had in mind? Important note: I haven't tested yet.

lib/strong_migrations.rb
```diff
@@ -18,8 +18,13 @@ class << self
 the entire table to be rewritten.
 
 Instead, add the column without a default value,
-then change the default.
+then change the default, then make sure to backfill
+the data of existing rows in another migration.
 
+# First migration
+class AddSomeColumnToUsers < #{
+  ActiveRecord::VERSION::MAJOR > 5 ? "ActiveRecord::Migration[#{ActiveRecord::VERSION::MAJOR}.#{ActiveRecord::VERSION::MINOR}]" : "ActiveRecord::Base"
```
ankane: Should still be `ActiveRecord::Migration` for ActiveRecord < 5.

lib/strong_migrations.rb
```diff
@@ -28,6 +33,25 @@ def up
...
+end
+
+# Second migration
+class BackfillSomeColumn < #{
+  ActiveRecord::VERSION::MAJOR > 5 ? "ActiveRecord::Migration[#{ActiveRecord::VERSION::MAJOR}.#{ActiveRecord::VERSION::MINOR}]" : "ActiveRecord::Base"
+}
+  disable_ddl_transaction!
+
+  def change
+    # Rails 5+
```
ankane: This as well.

ankane:
Cleaning up stale PRs

PR #50, opened 2018-10-09
Consider change_column migration to change integer limit as a safe operation in Postgres

manojmj92:
From `Postgres 9.2` [release notes](https://www.postgresql.org/docs/9.2/static/release-9-2.html):

```
Increasing the length limit for a varchar or varbit column, or removing the limit altogether, no longer requires a table rewrite. 

Similarly, increasing the allowable precision of a numeric column, or changing a column from constrained numeric to unconstrained numeric, no longer requires a table rewrite. 

Table rewrites are also avoided in similar cases involving the interval, timestamp, and timestamptz types.
```
At our company we use Postgres 9.6 and I was trying to write a migration that changes `integer` limit on a column from `4` to `8`, but `strong_migrations` raised error as this is assumed to be an unsafe migration because of `change_column`.
screen shot 2018-10-09 at 11 18 30 am

However, as it is clear from the release notes, it is safe to change integer limits from `Postgres 9.2+ `onwards, so I have made necessary changes to **not** raise error when such a migration is written.

Also, as per Postgres [release notes](https://www.postgresql.org/support/versioning/), EOL is now on verison 9.2 so I have made the assumption that Postgres version is 9.2+.

### DIFF
--- a/lib/strong_migrations/migration.rb
+++ b/lib/strong_migrations/migration.rb
@@ -52,10 +52,15 @@ module StrongMigrations
           end
         when :change_column
           safe = false
-          # assume Postgres 9.1+ since previous versions are EOL
-          if postgresql? && args[2].to_s == "text"
-            column = connection.columns(args[0]).find { |c| c.name.to_s == args[1].to_s }
-            safe = column && column.type == :string
+          # assume Postgres 9.3+ since previous versions are EOL
+          if postgresql?
+            if args[2].to_s == "text"
+              column = connection.columns(args[0]).find { |c| c.name.to_s == args[1].to_s }
+              safe = column && column.type == :string
+            elsif args[2].to_s == "integer" && !args[3][:limit].nil?
+              column = connection.columns(args[0]).find { |c| c.name.to_s == args[1].to_s }
+              safe = column && column.type == :integer
+            end
           end
           raise_error :change_column unless safe
         when :create_table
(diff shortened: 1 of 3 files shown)


waclock:
I've stumbled upon this same problem at least twice. Any reason why this wasn't merged onto master?

ankane:
Not sure why the PR was closed by the author the same day, but changing the size of an integer column is not safe.

waclock:
So postgres' documentation is not accurate? Or am I missing something?

ankane:
It seems like you haven't actually tried it. You can test this with:

```sql
CREATE TABLE t1 (value int);
INSERT INTO t1 SELECT i FROM generate_series(1,10000000) i;
ALTER TABLE t1 ALTER COLUMN value TYPE bigint; -- slow
```

The docs are referring to the `numeric` type, which is a variable length type, unlike `integer` and `bigint`.

```sql
CREATE TABLE t2 (value numeric(10, 1));
INSERT INTO t2 SELECT i FROM generate_series(1,10000000) i;
ALTER TABLE t2 ALTER COLUMN value TYPE numeric(20, 1); -- fast (precision increased, same scale)
ALTER TABLE t2 ALTER COLUMN value TYPE numeric; -- fast (unconstrained)
```

waclock:
I actually had tried it, and it was indeed slow. Thanks for clarifying!

ankane:
The table is locked during this period, which is what makes it unsafe. You can test this by trying to run any query against it.

PR #55, opened 2018-11-19
Add round brackets to be more consistent

hubertjakubiak:
I'm tired of updating migration when I see round brackets for `where` part and not for `update_all` method :)


ankane:
Thanks @hubertjakubiak, but I prefer the current style. The brackets around the `where` part are necessary.

PR #61, opened 2019-02-04
allow add comment changes to otherwise unchanged columns

AGarrow:
It is my understanding that adding a column annotation (comment) in postgres should be a safe migration. This PR allows that.

Please let me know if I missed anything, or if my understanding is off. Happy to make changes as need be and assist in maintaining this change.

Also, I added a db cleaner to run between tests since I was having some errors thrown from duplicate columns.

Thanks,
Alexi

### DIFF
--- a/lib/strong_migrations/migration.rb
+++ b/lib/strong_migrations/migration.rb
@@ -63,7 +63,6 @@ module StrongMigrations
           table, column, type, options = args
           options ||= {}
           default = options[:default]
-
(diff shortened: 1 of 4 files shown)


ankane:
Hey @AGarrow, thanks for the PR 👍 Rails still runs an alter column command, so I'm not sure about the safety. Also, I don't think this accounts for all cases that could cause issues, like the precision, scale, or other properties changing. Since I don't think this is a common situation and it'll require some complexity to handle different cases, I don't think it's worth the effort to support this use case in the gem.

Issue #59, opened 2019-01-23
[Idea] Warn on the addition of a validated foreign key


iamvery:
I recently ran into a problem adding a new foreign key constraint at scale. It seems that by default, when adding a foreign key, the entire table is validated for the new constraint which performs some table locking.

From the [PostgreSQL docs](https://www.postgresql.org/docs/9.6/sql-altertable.html)

>If the constraint is marked NOT VALID, the potentially-lengthy initial check to verify that all rows in the table satisfy the constraint is skipped. 

In my particular case, I was adding the foreign key to a new column which would be NULL for all existing records anyway thus not needing the validation at all. Here's the full example that caused me problems:

```ruby
add_column :foos, :bar_id, :integer
add_foreign_key :foos, :bars
```

And here is the preferred way:

```ruby
add_column :foos, :bar_id, :integer
add_foreign_key :foos, :bars, validate: false
```
Note: `validate: false` which results in `NOT VALID` option being added to the produced `ALTER TABLE` statement.

What do you think?

ankane:
Hey @iamvery, thanks for the suggestion. I think this is best as a custom check. I imagine most of the time foreign keys need to be valid.

iamvery:
Thanks for the response @ankane! I'm a little indifferent on this, but I understand your point. I'm indifferent, because I feel like if the purpose of this library is to protect users from themselves it might go so far to recommend against typical cases like this in the name of safety.

>Strong Migrations detects potentially dangerous operations in migrations, prevents them from running by default, and provides instructions on safer ways to do what you want.

Either way, the custom checks are a great feature that we can take advantage of. Thank you!

iamvery:
Thanks for the follow up here! It's going to be a few days before I can dig in, but this sounds great 👍

ankane:
Going to close this out, but feel free to create a PR

PR #62, opened 2019-02-21
Add check for validated foreign key constraint

iamvery:
Following #59, this PR adds a new check to protect against the addition of a validated foreign key constraint which acquires an AccessExclusiveLock which may result in significant performance problems on large, active tables.

### DIFF
--- a/lib/strong_migrations.rb
+++ b/lib/strong_migrations.rb
@@ -157,10 +157,10 @@ class Backfill%{migration_name} < ActiveRecord::Migration%{migration_suffix}
 end",
 
   add_foreign_key:
-"A new foreign key is validated by default. This aquires an AccessExclusiveLock
+"A new foreign key is validated by default. This acquires an AccessExclusiveLock
 which is expensive on large tables resulting in downtime. Instead, create an
 \"invalid\" foreign key constraint which may be explicitly validated later if
-necessary. The separate validation aquires a more agreeable RowShareLock.
+necessary. The separate validation acquires a more agreeable RowShareLock.
 
 class %{migration_name} < ActiveRecord::Migration%{migration_suffix}
   def change
--- a/lib/strong_migrations.rb
+++ b/lib/strong_migrations.rb
@@ -155,6 +155,18 @@ class Backfill%{migration_name} < ActiveRecord::Migration%{migration_suffix}
     %{code}
   end
 end",
+
+  add_foreign_key:
+"A new foreign key is validated by default. This aquires an AccessExclusiveLock
+which is expensive on large tables resulting in downtime. Instead, create an
+\"invalid\" foreign key constraint which may be explicitly validated later if
+necessary. The separate validation aquires a more agreeable RowShareLock.
+
+class %{migration_name} < ActiveRecord::Migration%{migration_suffix}
+  def change
+    %{code}
+  end
+end",
   }
 
   def self.add_check(&block)
--- a/lib/strong_migrations.rb
+++ b/lib/strong_migrations.rb
@@ -154,7 +154,7 @@ class Backfill%{migration_name} < ActiveRecord::Migration%{migration_suffix}
   def change
     %{code}
   end
-end"
+end",
   }
 
   def self.add_check(&block)
--- a/lib/strong_migrations.rb
+++ b/lib/strong_migrations.rb
@@ -164,7 +164,13 @@ necessary. The separate validation acquires a more agreeable RowShareLock.
 
 class %{migration_name} < ActiveRecord::Migration%{migration_suffix}
   def change
-    %{code}
+    %{add_foreign_key_code}
+  end
+end
+
+class ValidateForeignKey < ActiveRecord::Migration%{migration_suffix}
+  def change
+    %{validate_foreign_key_code}
   end
 end",
   }
--- a/lib/strong_migrations/migration.rb
+++ b/lib/strong_migrations/migration.rb
@@ -148,8 +148,8 @@ end"
 
           if postgresql? && validated
             raise_error :add_foreign_key,
-              add_foreign_key_code: foreign_key_code(from_table, to_table),
-              validate_foreign_key_code: validation_code(from_table, to_table)
+              add_foreign_key_code: command_str(:add_foreign_key, [from_table, to_table, validate: false]),
+              validate_foreign_key_code: command_str(:validate_foreign_key, [from_table, to_table])
           end
         end
 
@@ -232,18 +232,6 @@ end"
     def manual_validate_foreign_key_code(from_table, to_table)
       <<~CODE.strip
         ALTER TABLE #{from_table} VALIDATE CONSTRAINT fk_#{from_table}_#{to_table};
-      CODE
-    end
-
-    def foreign_key_code(from_table, to_table)
-      <<~CODE.strip
-        #{command_str(:add_foreign_key, [from_table, to_table, validate: false])}
-      CODE
-    end
-
-    def validation_code(from_table, to_table)
-      <<~CODE.strip
-        #{command_str(:validate_foreign_key, [from_table, to_table])}
       CODE
     end
 
--- a/lib/strong_migrations/migration.rb
+++ b/lib/strong_migrations/migration.rb
@@ -139,8 +139,7 @@ end"
           options ||= {}
           validated = options.fetch(:validate) { true }
 
-          five_point_two = Gem::Version.new("5.2")
-          if postgresql? && ActiveRecord.version < five_point_two
+          if postgresql? && ActiveRecord::VERSION::MAJOR > 5 || (ActiveRecord::VERSION::MAJOR == 5 && ActiveRecord::VERSION::MINOR >= 2)
             raise_error :add_foreign_key,
               add_foreign_key_code: manual_add_foreign_key_code(from_table, to_table),
               validate_foreign_key_code: manual_validate_foreign_key_code(from_table, to_table)
--- a/lib/strong_migrations/migration.rb
+++ b/lib/strong_migrations/migration.rb
@@ -141,7 +141,8 @@ end"
 
           if postgresql? && validated
             raise_error :add_foreign_key,
-              code: foreign_key_code(from_table, to_table)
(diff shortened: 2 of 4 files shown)


lib/strong_migrations.rb
```diff
@@ -154,7 +154,19 @@ class Backfill%{migration_name} < ActiveRecord::Migration%{migration_suffix}
   def change
     %{code}
   end
-end"
+end",
+
+  add_foreign_key:
+"A new foreign key is validated by default. This aquires an AccessExclusiveLock
```
iamvery: typo: "acquires"

lib/strong_migrations.rb
```diff
@@ -154,7 +154,19 @@ class Backfill%{migration_name} < ActiveRecord::Migration%{migration_suffix}
...
+which is expensive on large tables resulting in downtime. Instead, create an
+\"invalid\" foreign key constraint which may be explicitly validated later if
+necessary. The separate validation aquires a more agreeable RowShareLock.
```
iamvery: typo: "acquires"

lib/strong_migrations/migration.rb
```diff
@@ -186,6 +195,24 @@ def backfill_code(table, column, default)
...
+        <<~CODE.strip
+          #{command_str(:add_foreign_key, [from_table, to_table, validate: false])}
+              #{command_str(:validate_foreign_key, [from_table, to_table])}
+        CODE
+      else
+        <<~CODE.strip
+          safety_assured {
+                execute <<~SQL
+                  ALTER TABLE #{from_table} ADD CONSTRAINT fk_#{from_table}_#{to_table} FOREIGN KEY (#{to_table.to_s.singularize}_id) REFERENCES #{to_table} (id) NOT VALID;
+                  ALTER TABLE #{from_table} VALIDATE CONSTRAINT fk_#{from_table}_#{to_table};
```
iamvery: While reflecting on this solution, I realized that it probably isn't necessary to output the solution for old versions of Rails, because this problem itself cannot be produced on old versions since the `:validate` option didn't exist. Do we agree on that?

lib/strong_migrations/migration.rb
```diff
@@ -186,6 +195,24 @@ def backfill_code(table, column, default)
       end
     end
 
+    def foreign_key_code(from_table, to_table)
+      if ActiveRecord::VERSION::MAJOR > 5 || ActiveRecord::VERSION::MAJOR == 5 && ActiveRecord::VERSION::MINOR >= 2
+        <<~CODE.strip
+          #{command_str(:add_foreign_key, [from_table, to_table, validate: false])}
+              #{command_str(:validate_foreign_key, [from_table, to_table])}
```
iamvery: I'm not certain that it's safe to validate the constraint in the same transaction as the foreign key being added. I'm still learning about these aspects of database management. Do you have any insight?

test/strong_migrations_test.rb
```diff
@@ -213,12 +213,18 @@ def version
...
-class Custom < TestMigration
+class AddForeignKey < TestMigration
   def change
     add_foreign_key :users, :other
   end
 end
 
+class Custom < TestMigration
+  def change
+    add_foreign_key :users, :other, validate: false
```
iamvery: I ended up getting away with using a similar example for the "Custom" scenario since built-in checks are run first.

lib/strong_migrations/migration.rb
```diff
@@ -119,6 +119,15 @@ def method_missing(method, *args, &block)
             raise_error :change_column_null,
               code: backfill_code(table, column, default)
           end
+        when :add_foreign_key
+          from_table, to_table, options = args
+          options ||= Hash.new
```
ankane: `{}`

lib/strong_migrations/migration.rb
ankane: I'm pretty sure they need to be in separate transactions.

lib/strong_migrations/migration.rb
```diff
@@ -119,6 +119,15 @@ def method_missing(method, *args, &block)
...
+          validated = options.fetch(:validate) { true }
+
+          if validated
```
ankane: afaik, this only applies to Postgres.

lib/strong_migrations/migration.rb
ankane: Older versions of Rails behave as `validate: true`, which is unsafe, so think this is needed. Also need a down migration.

lib/strong_migrations/migration.rb
```diff
@@ -186,6 +195,24 @@ def backfill_code(table, column, default)
       end
     end
 
+    def foreign_key_code(from_table, to_table)
+      if ActiveRecord::VERSION::MAJOR > 5 || ActiveRecord::VERSION::MAJOR == 5 && ActiveRecord::VERSION::MINOR >= 2
```
ankane: Parentheses on the 2nd part would be good for readability.

test/strong_migrations_test.rb
ankane: Would be good to change the custom example to something totally different, like `add_column` with a specific name. I can do this afterwards though.

iamvery: If that's the case, then I would also want to always mark this method as unsafe for older versions, right?

test/strong_migrations_test.rb
```diff
@@ -368,6 +374,10 @@ def test_down
     assert_safe SafeUp, direction: :down
   end
 
+  def test_add_foreign_key
+    assert_safe AddForeignKey
```
iamvery: Temporarily flipped this to `assert_safe` so I can inspect the "suggested change" messaging. Should flip this back before calling the work here done.

TODO
```diff
@@ -0,0 +1 @@
+read: https://travisofthenorth.com/blog/2017/2/2/postgres-adding-foreign-keys-with-zero-downtime
```
iamvery: This file was probably added by accident in the last WIP commit. Will get it out when finalizing the work here.

lib/strong_migrations/migration.rb
```diff
@@ -119,6 +119,16 @@ def method_missing(method, *args, &block)
             raise_error :change_column_null,
               code: backfill_code(table, column, default)
           end
+        when :add_foreign_key
```
iamvery: I'm thinking we'll have to reject any call to this method on Rails < 5.2 since `validate: true` is implied.

lib/strong_migrations/migration.rb
```diff
@@ -119,6 +119,23 @@ def method_missing(method, *args, &block)
...
+          from_table, to_table, options = args
+          options ||= {}
+          validated = options.fetch(:validate) { true }
+
+          five_point_two = Gem::Version.new("5.2")
+          if postgresql? && ActiveRecord.version < five_point_two
```
iamvery: I feel like this is a better way to do gem version comparisons than boolean operations on `ActiveRecord::VERSION`

iamvery:
@ankane finally found some time to update this PR. Let me know what you think!

test/test_helper.rb
```diff
@@ -50,8 +50,8 @@ def postgres?
 end
 
 StrongMigrations.add_check do |method, args|
-  if method == :add_foreign_key && args[0].to_s == "users"
-    stop! "No foreign keys on the users table"
+  if method == :add_column && args[1].to_s == "forbidden"
+    stop! "Cannot add forbidden column"
```
iamvery: Went ahead and changed the custom example to keep the test suite happy.

test/strong_migrations_test.rb
```diff
@@ -368,8 +374,13 @@ def test_down
     assert_safe SafeUp, direction: :down
   end
 
+  def test_add_foreign_key
+    skip unless postgres?
```
iamvery: It has been a really long time since I've used mysql. How confident are we that the unsafe foreign key does not apply to it?

ankane: I'm not sure if it's safe with MySQL, it's fine to start with just Postgres.

test/test_helper.rb
ankane: 👍

lib/strong_migrations/migration.rb
ankane: Please keep it consistent for now.

lib/strong_migrations/migration.rb
```diff
@@ -186,6 +203,30 @@ def backfill_code(table, column, default)
...
+
+    def foreign_key_code(from_table, to_table)
+      <<~CODE.strip
+        #{command_str(:add_foreign_key, [from_table, to_table, validate: false])}
+      CODE
+    end
+
+    def validation_code(from_table, to_table)
+      <<~CODE.strip
+        #{command_str(:validate_foreign_key, [from_table, to_table])}
```
ankane: The interpolation shouldn't be needed. Since all of the commands are one-liners, I think we can get rid of the helper methods as well.

lib/strong_migrations/migration.rb
ankane: Makes sense.

lib/strong_migrations/migration.rb
iamvery: See #74

lib/strong_migrations/migration.rb
```diff
@@ -134,6 +134,22 @@ def change
             raise_error :change_column_null,
               code: backfill_code(table, column, default)
           end
+        when :add_foreign_key
+          from_table, to_table, options = args
+          options ||= {}
+          validated = options.fetch(:validate) { true }
+
+          if postgresql? && ActiveRecord::VERSION::MAJOR > 5 || (ActiveRecord::VERSION::MAJOR == 5 && ActiveRecord::VERSION::MINOR >= 2)
```
iamvery: Fwiw, this would be much simpler by comparing Gem::Version objects. See #74

ankane:
Thanks @iamvery, merged in https://github.com/ankane/strong_migrations/commit/cf25a6c70934b3a383feed707a360675ee61ac29.

PR #67, opened 2019-04-17
readme points out mysql/postgres differences

michaelglass:
adding this because, e.g. while adding a nullable column is cheap in Postgres, it is not, as far as I can tell, in mysql.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Strong Migrations detects potentially dangerous operations in migrations, preven
 
 ## Dangerous Operations
 
-The following operations can cause downtime or errors:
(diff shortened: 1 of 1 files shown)


ankane:
Hey @michaelglass, thanks for the PR 👍 Most of my experience is with Postgres, but if there are dangerous operations for MySQL that aren't being marked as dangerous, I think it'd be best to add them to the project. I'm having a hard time telling what's dangerous from the article.

michaelglass:
I'm happy to expand upon the readme but I don't have the bandwidth right now to fix the gem. Thoughts?

ankane:
I think the most useful thing would be to create a list of those operations here (including migration syntax). Then I can add them once it's confirmed they're dangerous.

ankane:
Cleaning up stale issues

PR #68, opened 2019-04-17
Recommend installing in development only

samjewell:
Update the Readme installation Gemfile code

This change makes the gemfile example code with the rest of the readme: specifically the very start:
> Catch unsafe migrations at dev time

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Catch unsafe migrations at dev time
 Add this line to your application’s Gemfile:
 
 ```ruby
-gem 'strong_migrations'
+gem "strong_migrations", group: :development
 ```
 
 ## How It Works


ankane:
Hey @samjewell, thanks for the PR, but the gem is needed in all environments.

bheemreddy181:
@ankane Can I know the benefits of having this gem in other environments other than dev - 
The one I know is auto_analyze feature

ankane:
@bheemreddy181 try running migrations that have safety_assured without the gem in production.

samjewell:
@ankane Thanks for the `safety_assured` explanation.

Given that the gem is need in all environments, how would you feel about changing the very first lines of the Readme, from:
> Catch unsafe migrations in development

To something like:
> Catch unsafe migrations before they run

?

ankane:
@samjewell I think the wording is fine as is. The goal is to still catch migrations in development.

PR #74, opened 2019-05-17
Use version object

iamvery:
The current mechanism for checking ActiveRecord version has worked well to date. However in [#62](https://github.com/ankane/strong_migrations/pull/62/files#r270973854), we came up on a need for making some code conditional on a more specific version of the library. It's quite a bit easier to do this by comparing `Gem::Version` objects, but I got the feedback to keep things consistent.

So I'm proposing we change how things are done everywhere, so I can use `Gem::Version` in that PR 😄


ankane:
Hey @iamvery, thanks for the PR, but the next version of Strong Migration will only support Rails 5+, so all of these checks will be removed. See `rails6` branch.

iamvery:
Wonderful!

Issue #75, opened 2019-05-29
[Idea] Allow setting postgresql_version in config


viamin:
Add a `target_postgresql_version` (or something like that) to StrongMigrations config. 
i.e. `StrongMigrations.target_postgres_version = 10.7` which is checked in postgresql_version https://github.com/ankane/strong_migrations/blob/981d4f02e4c758238adf26451b9927270667344c/lib/strong_migrations/checker.rb#L194

Background: 
Adding a column with a default is unsafe in PostgreSQL < 11, and will raise an error. This passed on my local machine because I'm running PostgreSQL v11.3. However, on our staging/production servers, we're running PostgreSQL 10.7. So strong_migrations let the migration through in development, but failed when building on staging. 

It would be safer to allow setting a production target version in StrongMigrations config so this doesn't slip through in development.

ankane:
Hey @viamin, sounds good to me if you want to submit a PR. I think it should only trust the target version in development and test environments. Rails 6 will introduce multiple databases (which could have different versions), but don't think we have to handle that case right now (unless you have ideas).

ankane:
Added in #76

PR #76, opened 2019-05-30
Allow setting StrongMigrations.target_postgresql_version

viamin:
Allows setting `StrongMigrations.target_postgresql_version` in an initializer that will override the local `postgresql_version` check. 

As described in https://github.com/ankane/strong_migrations/issues/75

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,22 +61,22 @@ end
 
 1. Tell ActiveRecord to ignore the column from its cache
 
-  ```ruby
-  class User < ApplicationRecord
-    self.ignored_columns = ["some_column"]
-  end
-  ```
+```ruby
+class User < ApplicationRecord
+  self.ignored_columns = ["some_column"]
+end
+```
 
 2. Deploy code
 3. Write a migration to remove the column (wrap in `safety_assured` block)
 
-  ```ruby
-  class RemoveSomeColumnFromUsers < ActiveRecord::Migration[5.2]
-    def change
-      safety_assured { remove_column :users, :some_column }
-    end
+```ruby
+class RemoveSomeColumnFromUsers < ActiveRecord::Migration[5.2]
+  def change
+    safety_assured { remove_column :users, :some_column }
   end
-  ```
+end
+```
 
 4. Deploy and run migration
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -61,22 +61,22 @@ end
 
 1. Tell ActiveRecord to ignore the column from its cache
 
-  ```ruby
```
ankane: Please keep the changeset to a minimum.

lib/strong_migrations.rb
```diff
@@ -9,11 +9,12 @@
 
 module StrongMigrations
   class << self
-    attr_accessor :auto_analyze, :start_after, :checks, :error_messages
+    attr_accessor :auto_analyze, :start_after, :checks, :error_messages, :target_postgresql_version
   end
   self.auto_analyze = false
   self.start_after = 0
   self.checks = []
+  self.target_postgresql_version = nil
```
ankane: Variables are `nil` by default.

lib/strong_migrations/version.rb
```diff
@@ -1,3 +1,3 @@
 module StrongMigrations
-  VERSION = "0.4.0"
+  VERSION = "0.4.1"
```
ankane: It's best to leave versioning to maintainers.

lib/strong_migrations/checker.rb
```diff
@@ -192,7 +192,34 @@ def postgresql?
...
+
+    def rails_test_or_development?
+      if defined?(Rails)
+        Rails.env.test? || Rails.env.development?
+      else
+        false
+      end
+    end
+
+    def target_version
```
ankane: Think we can simplify this method significantly. Basically, just need:

```ruby
StrongMigrations.target_postgresql_version.to_i * 10000
```

Since there should only be changes to safety between major versions.

lib/strong_migrations/version.rb
viamin: Old habits...

README.md
viamin: I think that was due to my editor settings. I've reverted that portion of the change.

lib/strong_migrations/checker.rb
viamin: Done. Updated the README as well.

ankane:
Looks great, thanks @viamin 👍

PR #78, opened 2019-06-18
Suggest to use temporary model for the backfilling [skip ci]

sharshenov:
The migrations should not be tied to an app code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ See the next section for how to backfill.
 
 #### Bad
 
-Backfilling in the same transaction that alters a table locks the table for the [duration of the backfill](https://wework.github.io/data/2015/11/05/add-columns-with-default-values-to-large-tables-in-rails-postgres/).
+Backfilling in the same transaction that alters a table locks the table for the [duration of the backfill](https://wework.github.io/data/2015/11/05/add-columns-with-default-values-to-large-tables-in-rails-postgres/). Reffering to a model that may be changed (e.g. renamed) later is also a possible point of failure.
 
 ```ruby
 class AddSomeColumnToUsers < ActiveRecord::Migration[5.2]
@@ -134,14 +134,18 @@ Also, running a single query to update data can cause issues for large tables.
 
 #### Good
 
-There are three keys: batching, throttling, and running it outside a transaction. Use the Rails console or a separate migration with `disable_ddl_transaction!`.
+There are three keys: batching, throttling, and running it outside a transaction. Use the Rails console or a separate migration with `disable_ddl_transaction!` and temporary model. 
 
 ```ruby
 class BackfillSomeColumn < ActiveRecord::Migration[5.2]
   disable_ddl_transaction!
+  
+  class TmpUser < ActiveRecord::Base
+    self.table_name = 'users'
+  end
 
   def change
-    User.unscoped.in_batches do |relation|
+    TmpUser.unscoped.in_batches do |relation|
       relation.update_all some_column: "default_value"
       sleep(0.1) # throttle
     end


ankane:
Hey @sharshenov, thanks for the PR 👍 I’ve heard this advice before, but don’t agree with it.

sharshenov:
@ankane Could you share your arguments? I'm not trying to start a holy-war. Just curious.

ankane:
Sure, maybe you can explain the reasons that "migrations should not be tied to an app code" and we can go from there.

sharshenov:
OK. But before I start, I would like to say thank you for the great projects (not only strong_migrations, but `searchkick`, `groupdate`, `pghero`). I'm using them very often.

From my experience, migrations that rely on app models or other "external code" can behave differently through the time. Here is a made up scenario:

DB schema:

```ruby
ActiveRecord::Schema.define(version: 2019_06_21_164528) do

  enable_extension "plpgsql"

  create_table "users", force: :cascade do |t|
    t.string "name"
    t.datetime "created_at", null: false
    t.datetime "updated_at", null: false
  end

end

```

And user model:

```ruby
class User < ApplicationRecord
end

```

Creating `first_name` && `last_name` columns to fill them up later:

```ruby
class AddFirstNameAndLastNameToUsers < ActiveRecord::Migration[5.2]
  def change
    add_column :users, :first_name, :string
    add_column :users, :last_name, :string
  end
end
```

Migrating data with a migration:
```ruby
class MigrateUserNames < ActiveRecord::Migration[5.2]
  def up
    User.unscoped.find_each do |u|
      u.first_name, u.last_name = u.name.split(' ', 2)
      u.save!
    end
  end

  def down
    User.unscoped.find_each do |u|
      u.name = [u.first_name, u.last_name].join(' ')
      u.save!
    end
  end
end
```

And dropping `name` afterwards:
```ruby
class DropNameInUsers < ActiveRecord::Migration[5.2]
  def change
    remove_column :users, :name, :string
  end
end
```

At this point everyting works correct... until we rename the table and the model.

```ruby
class Client < ApplicationRecord
end
```

```ruby
class RenameUsersToClients < ActiveRecord::Migration[5.2]
  def change
    rename_table :users, :clients
  end
end
```

Your teammate returns from a vacation, pulls master and runs the migrations. The migration that refers to the `User` model will raise `uninitialized constant MigrateUserNames::User`. To make it work, your teammate has to run the migration by switching to a proper commit.

A temporary model definition makes the migration isolated.

```ruby
class MigrateUserNames < ActiveRecord::Migration[5.2]
  class TmpUser < ActiveRecord::Base
    self.table_name = 'users'
  end

  def up
    TmpUser.unscoped.find_each do |u|
      u.first_name, u.last_name = u.name.split(' ', 2)
      u.save!
    end
  end

  def down
    TmpUser.unscoped.find_each do |u|
      u.name = [u.first_name, u.last_name].join(' ')
      u.save!
    end
  end
end
```

This solution doesn't play nice with STI (in that case I would prefer pure-sql snippet).

januszm:
@sharshenov  I believe that the problems that you described can be solved by:
1. not using migrations to change data (e.g. use scripts instead)
2. using `rails db:schema:load` / `rails db:structure:load` rather than migrations in development environments.

sharshenov:
@januszm I agree with you at 1st point. But this PR is for the README where migration [is already used for backfiling data](https://github.com/ankane/strong_migrations/blob/f0a873e1c56fd00d1436ce829d480833dfe870d4/README.md#bad-2).
Regarding the 2nd, it may be an option for development environment only, but not for production.

ankane:
Thanks for the detailed explanation (it matches the reasoning for this approach I've heard before). In my experience, it's rare that models get renamed, and if someone on the team goes on vacation and falls too far behind on migrations, they can do a schema load like new hires need to do.

sharshenov:
I agree that the described case is not common thing. Closing PR in order not to flood

PR #80, opened 2019-06-18
link to discussion about 4-column indexes

jjb:
😄


ankane:
Not sure the conversation explains it well. Would be nice to find an article that discusses multicolumn indexes.

ankane:
Hey @jjb, updated the explanation. Let me know if it could still be improved.

PR #86, opened 2019-08-22
Add `remove_index` checker

kukula:
We need this check because removing indexes locks the table as well. And it can be dangerous.
Thank you for the gem! It's very helpful.


arvinwiyono:
Hi @kukula, you might want to update the README to include `remove_index`

ankane:
Hey @kukula, thanks for the PR, but I haven't experienced issues with remove index at scale, so I don't think it makes sense to add. See #8, #22, and #33.

Feel free to use a custom check in your own app.

kukula:
Thanks for quick reply @ankane 
Yeap, we are using the fork of this repo.

PR #87, opened 2019-08-23
Support and encourage adding concurrently as an option to add_reference

mlarraz:
* This is preferable since it avoids the extra line
* This is not currently detected and triggers a false positive

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,8 +202,7 @@ class AddReferenceToUsers < ActiveRecord::Migration[5.2]
   disable_ddl_transaction!
 
   def change
-    add_reference :users, :city, index: false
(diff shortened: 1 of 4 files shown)


ankane:
Beautiful, thanks @mlarraz 👍

fsateler:
`add_reference` is dangerous due to #125: it runs multiple statements. I agree that detecting multiple statements is difficult and thus I understand why that issue was closed. Nevertheless, I don't think the documentation should suggest dangerous "fixes".

ankane:
Hey @fsateler, having multiple statements with `disable_ddl_transaction!` isn't dangerous. Any migration that creates indexes concurrently can be partially applied and will need manually resolved if index creation fails (which isn't dangerous).

PR #92, opened 2019-10-03
Add check for deleting an index before creating the new one

bzf:
When replacing an index in the database it's easy to not think about
the time between the deleting of the old index and creating the new one,
especially if it's on a large table. If it's a heavily used table, not
having an index will make the database spend a lot of time running those
queries until the new index is in place. This might break your web
server since most requests will timeout waiting on these queries to
complete and block other requests from coming through.

This adds a check to see if you're dropping the old index before you've
created the new one recommends creating the new index before dropping
the old one.

EDIT: Also, thanks for this awesome gem 😄

### DIFF
--- a/lib/strong_migrations/checker.rb
+++ b/lib/strong_migrations/checker.rb
@@ -173,6 +173,13 @@ end"
         end
       end
 
+      remove_index_index = @ordered_migrations.index { |k| k.first == :remove_index }
+      add_index_index = @ordered_migrations.drop(remove_index_index).index { |k| k.first == :add_index } if remove_index_index.present?
+
+      if remove_index_index.present? && add_index_index
+        raise_error :dangerous_replace_index
+      end
+
       result = yield
 
       if StrongMigrations.auto_analyze && direction == :up && postgresql? && method == :add_index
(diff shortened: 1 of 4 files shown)


lib/strong_migrations/checker.rb
```diff
@@ -170,6 +173,13 @@ def change
         end
       end
 
+      remove_index_index = @ordered_migrations.index { |k| k.first == :remove_index }
+      add_index_index = @ordered_migrations.drop(remove_index_index).index { |k| k.first == :add_index } if remove_index_index.present?
+
+      if remove_index_index.present? && add_index_index
+        raise_error :dangerous_replace_index
+      end
```
bzf: I wasn't so sure where to put this check, so I just added it here for discussion. It would be nice if there was a class for each "checker" that could raise error and have its own methods for this sort of stuff.

Also, currently it doesn't check if you're removing and adding the index for the same table, do you think that's something that should be added?

Do you have any suggestion here?

ankane:
Hey @bzf, thanks for the PR, including tests 💯 I'm a bit hesitant to add since I think it'll cause a lot of false positives. As you mention, the check can be scoped to the same table (or even covered indexes), but there could still be existing indexes on the table that make the migration safe. We could check for that as well, but it feels like a lot of complexity.

ankane:
Cleaning up stale PRs

PR #93, opened 2019-10-30
Add `AddTypeColumn` checker

okuramasafumi:
`type` column is reserved by Rails and is used for STI.
Soon after `type` column is added, Rails looks for a subclass according to `type` column. This might cause downtime.

### DIFF
--- a/lib/strong_migrations.rb
+++ b/lib/strong_migrations.rb
@@ -44,6 +44,9 @@ end%{append}",
 "There's no equality operator for the json column type, which
 causes issues for SELECT DISTINCT queries. Use jsonb instead.",
 
+    add_type_column:
+"Adding type column in migration might cause downtime.",
+
     change_column:
 "Changing the type of an existing column requires the entire
 table and indexes to be rewritten. A safer approach is to:
--- a/lib/strong_migrations/checker.rb
+++ b/lib/strong_migrations/checker.rb
@@ -93,6 +93,10 @@ end"
           if type.to_s == "json" && postgresql?
             raise_error :add_column_json
           end
+
+          if column == ActiveRecord::Base.inheritance_column.to_sym
+            raise_error :add_type_column
+          end
         when :change_column
           table, column, type = args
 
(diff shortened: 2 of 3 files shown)


ankane:
Hey @okuramasafumi, thanks for the PR 👍 From my testing, this should only cause issues if a default value is specified that doesn't match up with a subclass. I'm a bit hesitant to add since the primary goal of the project is to alerts developers of migrations that appear to run fine in development but cause issues in a production environment, and this should really only cause issues if a developer doesn't test at all, so I think it'll lead to a lot of false positives.

okuramasafumi:
@ankane Thanks for the quick reply. This check is useful when deployment takes long, according to a case from a friend of mine (in his case, it took about 40 minutes and caused downtime).
I don't know if we should take care of the cases depending on how long deployment takes. If you think it's not a thing we should take care of, please close this issue!

ankane:
Hey @okuramasafumi, it's hard to know what caused the downtime in that specific instance without knowing more details, but from my tests, adding a `type` column alone won't cause it. If you have more details, feel free to share. Otherwise, we can close this out.

andyatkinson:
@okuramasafumi Were they using PostgreSQL? As @ankane hinted at, I would guess that it wasn't just the `type` column itself, but that the column was being added with a default. According to this handy [Safe Operations For High Volume PostgreSQL](https://www.braintreepayments.com/blog/safe-operations-for-high-volume-postgresql/) both adding a column and setting a default is not recommended. Instead they recommend the column is added and the default is applied as a second operation. If that was the issue, hope this information helps. I agree that I think having a check on just a `type` column would result in a lot of false positives and confusion.

yuki-ycino:
There was a problem in our environment.

Although a migration and ActiveRecord model subclass for STI were consistent, an old Rails application instance were up during migration for `type` column. It caused a downtime since migration and deployment don’t finish at the same time.

ankane:
Hey @yuki-ycino, thanks for the follow up. It sounds like this was app-specific, so it's probably best handled with a [custom check](https://github.com/ankane/strong_migrations#custom-checks).

PR #94, opened 2019-11-08
Allow foreign key additions to be safe in Postgres 11+

michael-groble:
In Postgres 11, adding a foreign key constraint requires a `SHARE ROW EXCLUSIVE` lock instead of an `ACCESS EXCLUSIVE` one ([docs](https://www.postgresql.org/docs/11/sql-altertable.html)).

It seems that when this check was added in #62, it was based on PG 10 or earlier?  Would appreciate any feedback.

### DIFF
--- a/lib/strong_migrations/checker.rb
+++ b/lib/strong_migrations/checker.rb
@@ -142,7 +142,7 @@ end"
           options ||= {}
           validate = options.fetch(:validate, true)
 
-          if postgresql?
+          if postgresql? && postgresql_version < 110000
             if ActiveRecord::VERSION::STRING >= "5.2"
               if validate
                 raise_error :add_foreign_key,
--- a/test/strong_migrations_test.rb
+++ b/test/strong_migrations_test.rb
@@ -393,26 +393,21 @@ class StrongMigrationsTest < Minitest::Test
   end
 
   def test_add_foreign_key
+    StrongMigrations.target_postgresql_version = 11
     assert_safe AddForeignKey
-  end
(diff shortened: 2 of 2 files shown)


ankane:
Hey @michael-groble, thanks for the PR! Need to look into this more when I have some time. The Postgres 11 and 12 docs still document the current approach as the way "to add a foreign key constraint to a table with the least impact on other work".

michael-groble:
Noting the `VALIDATE CONSTRAINT` command wants both a `SHARE UPDATE EXCLUSIVE` and `ROW SHARE` lock on the table.  Meaning the difference is that the "totally safe" way of a separate validate step allows updates/inserts/deletes while the default one does not (`ROW EXCLUSIVE` appears to be the only difference). So definitely different "degrees" of safeness in this one.

We're upgrading rails versions and can use the version of this gem that adds this check.  Trying to think through the appropriate use for us (absolute safety vs burden of changing how we do all foreign keys).  I'd be OK if you would rather keep things totally safe in the gem.  Possibly some updated documentation to point out the lower level of impact in 11+.

ankane:
Yeah, I think Strong Migrations should protect against blocking updates/inserts/deletes when possible.

From what I can tell, the lock change happened in Postgres 9.5 ([source](https://www.depesz.com/2015/04/23/waiting-for-9-5-reduce-lock-levels-of-some-trigger-ddl-and-add-fks/)). Will update the docs with the correct lock type.

Thanks for finding this.

PR #97, opened 2019-11-20
Use proper url for 'alphabetize them'

alvir:


ankane:
Thanks @alvir! (guessing the blog switched time zones)

PR #100, opened 2019-12-05
Add spacing to hash - styling improvement

hubertjakubiak:
According to https://github.com/rubocop-hq/ruby-style-guide#spaces-and-braces with spacing is better :)


ankane:
Hey @hubertjakubiak, thanks for the suggestion. As you can imagine, I prefer the current style.

PR #101, opened 2019-12-18
Fix migration class name when displaying hints

fatkodima:
Before
Screen Shot 2019-12-18 at 03 15 36


ankane:
Yikes, this bug has been around since the 0.4.1 release. Thanks @fatkodima for the fix and a test to make sure it doesn't break again. Will push out a new release shortly.

Issue #95, opened 2019-11-11
Ideas


ankane:
Ideas

- helper methods (could be nice for more complex operations) - similar to [GitLab](https://gitlab.com/gitlab-org/gitlab-foss/blob/master/lib%2Fgitlab%2Fdatabase%2Fmigration_helpers.rb)
- disable specific checks - `StrongMigrations.disable(:add_index)`
- optional checks - `StrongMigrations.enable(:remove_index)`
    - drop index concurrently
    - enforce timestamptz
    - reversible migrations
- add functionality similar to other gems
    - generate index - `rails g index`
    - set default lock timeout

I'm still on the fence about many of them. If anyone would like to discuss any of the ideas above, please create a new issue.

mvz:
I'd be very interested in some helper method for renaming tables.

fatkodima:
Is there really a need for `change_column_null_safely` after added `add_null_constraint_safely` / `remove_null_constraint_safely` ?

bheemreddy181:
Can we add a retry mechanism if migrations fail due to a timeout?

ankane:
Hey @bheemreddy181, it's an interesting idea.

For statements run outside a transaction, retrying the statement should work:

```ruby
class AddSomeIndexToUsers < ActiveRecord::Migration[6.0]
  disable_ddl_transaction!

  def change
    add_index :users, :some_column, algorithm: :concurrently
  end
end
```

For statements run inside a transaction, the entire transaction needs to be re-run. For migrations that only use Active Record's DDL transaction and have no side effects, retrying the entire migration should work:

```ruby
class SimpleMigration < ActiveRecord::Migration[6.0]
  def change
    add_column :users, :some_column, :string
  end
end
```

However, retrying more complex migrations can cause issues or errors (which may be an acceptable trade-off).

```ruby
class ComplexMigration < ActiveRecord::Migration[6.0]
  def change
    add_column :users, :some_column, :string
    commit_db_transaction # we can avoid retry if this method is called
    begin_db_transaction
    add_column :users, :other_column, :string
  end
end
```

Or

```ruby
class ComplexMigration2 < ActiveRecord::Migration[6.0]
  disable_ddl_transaction!

  def change
    add_column :users, :some_column, :string

    connection.transaction do
      add_column :users, :other_column, :string
    end
  end
end
```

[...]

PR #102, opened 2019-12-18
[WIP] Add migration helper methods [skip ci]

fatkodima:
Related #95 

Currently, I'm not asking for a thorough review, just POC validation.

This adds (documentation and usage examples provided inline with method definitions):
1. `add_foreign_key_concurrently`. If we can update `activerecord` dependency to `>= 5.2`, then this won't be needed as `add_foreign_key ..., validate: false` can be used directly 
2. `update_column_in_batches`
3. `add_column_with_default`
4. `rename_column_concurrently`, `undo_rename_column_concurrently`, `cleanup_after_rename_column_concurrently`, `undo_cleanup_after_rename_column_concurrently` - for working with renaming columns
5. `change_column_type_concurrently`, `cleanup_after_change_column_type_concurrently`

While writing this, I have figured it out, that it will also be useful to add something like `rename_table_concurrently`. If you have other ideas of what can be useful to add, please share.

Gitlab also has some code for *background* migrations for extra large tables. I do not considered them here to make this PR more digestible. Those will be added separately, if needed.

Todo:
- [ ] support mysql
- [ ] add tests
- [ ] update documentation

### DIFF
--- a/lib/strong_migrations.rb
+++ b/lib/strong_migrations.rb
@@ -3,6 +3,7 @@ require "active_support"
 require "strong_migrations/checker"
 require "strong_migrations/database_tasks"
 require "strong_migrations/migration"
+require "strong_migrations/migration_helpers"
 require "strong_migrations/railtie" if defined?(Rails)
 require "strong_migrations/unsafe_migration"
 require "strong_migrations/version"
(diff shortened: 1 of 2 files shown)


ankane:
Hey @fatkodima, thanks for another PR! Overall, I think migration helpers makes sense. A few notes:

1. Let's add them one at a time (or in logical groups) so we can discuss them individually
2. We shouldn't copy any code from GitLab

fatkodima:
@ankane 
In point 2, you mean due to licensing?
And what about mentioned changing `activerecord` requirement to `>= 5.2`?

ankane:
Yeah, due to licensing, and I'd rather recreate them from first principles in their simplest form to start.

Re activerecord: we shouldn't change the version requirements, but we can have some helpers only work with AR 5.2+ (and raise an error message for earlier versions) if it's significantly more work to make them work with earlier versions.

ankane:
Closing in favor of separate PRs. @fatkodima, would love to get your thoughts on #111 when you have a chance.

ankane:
Also, fwiw, I made the helpers opt-in on master to demonstrate what that would look like.

PR #103, opened 2019-12-19
Add add_foreign_key_concurrently migration helper

fatkodima:
This method is not needed for mysql. So should this be a no-op for mysql or is it enough that only postgres is mentioned for foreign keys section in readme?

### DIFF
--- a/lib/strong_migrations.rb
+++ b/lib/strong_migrations.rb
@@ -3,6 +3,7 @@ require "active_support"
 require "strong_migrations/checker"
 require "strong_migrations/database_tasks"
 require "strong_migrations/migration"
+require "strong_migrations/migration_helpers"
 require "strong_migrations/railtie" if defined?(Rails)
 require "strong_migrations/unsafe_migration"
 require "strong_migrations/version"
@@ -174,18 +175,15 @@ end",
 
     add_foreign_key:
 "New foreign keys are validated by default. This acquires an AccessExclusiveLock,
-which is expensive on large tables. Instead, validate it in a separate migration
+which is expensive on large tables. To work around this, you can use `add_foreign_key_concurrently`
+helper. This method ensures that no downtime is needed. It will create foreign key and validate it separately
 with a more agreeable RowShareLock.
 
 class %{migration_name} < ActiveRecord::Migration%{migration_suffix}
-  def change
-    %{add_foreign_key_code}
-  end
-end
+  disable_ddl_transaction!
 
-class Validate%{migration_name} < ActiveRecord::Migration%{migration_suffix}
   def change
-    %{validate_foreign_key_code}
+    %{command}
   end
 end",
   }
@@ -218,5 +216,6 @@ ActiveSupport.on_load(:active_record) do
 
   if defined?(ActiveRecord::Tasks::DatabaseTasks)
     ActiveRecord::Tasks::DatabaseTasks.singleton_class.prepend(StrongMigrations::DatabaseTasks)
+    ActiveRecord::Migration.include(StrongMigrations::MigrationHelpers)
   end
 end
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -227,48 +227,20 @@ end
 
 #### Good
 
-Instead, validate it in a separate migration with a more agreeable `RowShareLock`. This approach is documented by Postgres to have “[the least impact on other work](https://www.postgresql.org/docs/current/sql-altertable.html).”
```
ankane: The readme change looks great, but won't be able to update it until the new release is out. Can you back out this change and paste the updated markdown as a comment?

lib/strong_migrations.rb
```diff
@@ -218,5 +220,6 @@ def self.check_enabled?(check, version: nil)
 
   if defined?(ActiveRecord::Tasks::DatabaseTasks)
     ActiveRecord::Tasks::DatabaseTasks.singleton_class.prepend(StrongMigrations::DatabaseTasks)
+    ActiveRecord::Migration.send(:include, StrongMigrations::MigrationHelpers)
```
ankane: fwiw, `include` is public in all supported versions of Ruby.

lib/strong_migrations/migration_helpers.rb
```diff
@@ -0,0 +1,70 @@
...
+      else
+        options = foreign_key_options(from_table, to_table, options)
+
+        # Using NOT VALID constraint option allows to reduce the impact of adding
+        # a constraint on concurrent updates. With NOT VALID, the ADD CONSTRAINT
+        # command does not scan the table and can be committed immediately. Added
+        # constraint is enforced for any newly created data.
+        safety_assured do
+          execute <<~SQL
+            ALTER TABLE #{from_table}
```
ankane: Identifiers need to be quoted here and with the `VALIDATE CONSTRAINT` command - see the `constraint_str` method.

lib/strong_migrations/migration_helpers.rb
```diff
@@ -0,0 +1,70 @@
...
+    #
+    # See Rails' `add_foreign_key` for more info on the available options.
+    def add_foreign_key_concurrently(from_table, to_table, options = {})
+      ensure_not_in_transaction(__method__)
+
+      if ActiveRecord::VERSION::STRING >= "5.2"
+        add_foreign_key(from_table, to_table, options.merge(validate: false))
+        validate_foreign_key(from_table, to_table)
+      else
+        options = foreign_key_options(from_table, to_table, options)
```
ankane: Nice find!

lib/strong_migrations/migration_helpers.rb
```diff
@@ -0,0 +1,70 @@
...
+
+        # Using NOT VALID constraint option allows to reduce the impact of adding
```
ankane: I think we can remove all the comments.

ankane:
Almost forgot: for MySQL, let's add a `ensure_postgresql` method that raises an error for non-Postgres databases.

ankane:
Sorry, one more thought: we should be able to make this reversible so a separate down migration isn't needed.

README.md
fatkodima: To work around this, you can use `add_foreign_key_concurrently` helper method. It ensures that no downtime
is needed. It will create foreign key and validate it separately with a more agreeable `RowShareLock`.
This approach is documented by Postgres to have “[the least impact on other work](https://www.postgresql.org/docs/current/sql-altertable.html).”

```ruby
class AddForeignKeyOnUsers < ActiveRecord::Migration[6.0]
  def change
    add_foreign_key_concurrently :users, :orders
  end
end
```

fatkodima:
@ankane Thanks for review! 💪 💪  

Updated with your sugsestions.

ankane:
Awesome, looks great. Thanks @fatkodima!

I renamed the method to `add_foreign_key_safely` since it's more generic and therefore we can apply it to more methods (like `change_column_null_safely`, etc).

PR #104, opened 2019-12-19
Simpler 'adding a foreign key'

afrojas:
Since these are each single operations, isn't it possible to simply remove the transaction wrapping and perform both `ALTER TABLE` operations in the same migration?

If I'm missing something about what a transaction provides, we could still wrap each statement in its own transaction while keeping both operations in one migration, no?

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,17 +233,10 @@ For Rails 5.2+, use:
 
 ```ruby
 class AddForeignKeyOnUsers < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
+
   def change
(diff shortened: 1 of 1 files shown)


ankane:
Hey @afrojas, thanks for the PR. This is possible, but if a user forgets `disable_ddl_transaction!`, everything will appear fine in development but it'll lock the DB in production. Presenting it as two separate migrations is safer from this perspective. Once #103 is added, it'll be possible to do in a single migration (since we can check for the presence of a transaction).

ankane:
`add_foreign_key_safely` is now on master.

PR #105, opened 2019-12-21
Add add_column_safely and backfill_column_safely migration helpers

fatkodima:


lib/strong_migrations/migration_helpers.rb
```diff
@@ -73,16 +75,86 @@ def remove_null_constraint_safely(table_name, column_name, name: nil)
...
+        .order(table[primary_key].asc)
+        .take(1)
+
+      result = connection.exec_query(start_arel.to_sql)
+      return if result.empty?
+
+      start_pk = result.first[primary_key]
+
+      loop do
+        finish_arel = table
```
jturkel: Does this also need to add `.where(table[column_name].not_eq(value))` to avoid reprocessing rows during a retry? Same question for the actual update a few lines down.

fatkodima: I don't think so. This method updates rows in primary key order, so when restarting, the whole process will start from first (`start_pk`) unmatched row, which was last when update crashes, and move forward.

jturkel: That's true that the process will pickup where it left off before the restart but won't it unnecessarily update rows that were created after the column default was set but before the backfill process completes (i.e. rows were the DB filled in the column default on insert)? Might not be much of an issue in practice though.

fatkodima: > Might not be much of an issue in practice though.

Actually, this.

lib/strong_migrations/migration_helpers.rb
```diff
@@ -73,16 +75,86 @@ def remove_null_constraint_safely(table_name, column_name, name: nil)
...
+
+    def backfill_column_safely(table_name, column_name, value, batch_size: 1000)
+      ensure_not_in_transaction(__method__)
+
+      table = Arel::Table.new(table_name)
+      primary_key = connection.primary_key(table_name)
+
+      start_arel = table
+        .project(table[primary_key])
+        .where(table[column_name].not_eq(value))
```
ankane: Backfilling should only affect `NULL` columns. Would be good to include a test case for this situation (can do after initial PR if that's easier)

lib/strong_migrations.rb
```diff
@@ -25,28 +26,15 @@ class << self
   self.error_messages = {
     add_column_default:
 "Adding a column with a non-null default causes the entire table to be rewritten.
-Instead, add the column without a default value, then change the default.
+Instead, add the column without a default value, backfill and then change the default.
```
ankane: The default gets changed before backfilling.

lib/strong_migrations/migration_helpers.rb
```diff
@@ -73,16 +75,86 @@ def remove_null_constraint_safely(table_name, column_name, name: nil)
...
+          .where(table[primary_key].gteq(start_pk))
+
+        if finish_result
+          finish_pk = finish_result[primary_key]
+          update_arel = update_arel.where(table[primary_key].lt(finish_pk))
+          start_pk = finish_pk
+        end
+
+        safety_assured { execute(update_arel.to_sql) }
+
```
ankane: Need to throttle - `sleep(0.01)` (may need a global option for this later since more powerful databases may need less throttling, but let's deal with that later)

fatkodima:
@ankane Thanks for review! Updated with your suggestions.

fatkodima:
Updated with master.

fatkodima:
@ankane would you like making another round of reviews on this?

ankane:
Hey @fatkodima, sorry for the delay. Still thinking on #111.

ankane:

PR #106, opened 2019-12-23
Add rename_column_safely and rename_column_safely_cleanup migration helpers

fatkodima:
- [x] Tests

This depends on (and partially consists of) #105 changes, so that should be merged first.


ankane:

PR #107, opened 2019-12-29
Add change_column_safely migration helper

fatkodima:
This pr depends on (and consists of) changes from previous 2 prs, so should be merged after them.

### DIFF
--- a/lib/strong_migrations/migration_helpers.rb
+++ b/lib/strong_migrations/migration_helpers.rb
@@ -150,7 +150,7 @@ module StrongMigrations
       end
     end
 
-    def rename_column_safely(table_name, old, new)
+    def rename_column_safely(table_name, old, new, options = {})
       if !postgresql? && !mysql?
         raise StrongMigrations::Error, "`#{__method__}` is intended for Postgres and Mysql only"
       end
@@ -160,7 +160,7 @@ module StrongMigrations
 
       reversible do |dir|
         dir.up do
-          copy_column(table_name, old, new)
+          copy_column(table_name, old, new, options)
           safety_assured { create_column_rename_triggers(table_name, old, new) }
         end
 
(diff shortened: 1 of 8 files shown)


lib/strong_migrations/migration_helpers.rb
```diff
@@ -73,16 +75,172 @@ def remove_null_constraint_safely(table_name, column_name, name: nil)
...
+
+      default = options.delete(:default)
+
+      if postgresql_version >= 110000 || default.nil?
+        add_column(table_name, column_name, options)
+      else
+        reversible do |dir|
+          dir.up do
+            transaction do
+              add_column(table_name, column_name, type, default: nil, **options)
```
jturkel: Should this only add the column if it doesn't already exist to properly handle retries since the entire transaction isn't wrapped in a transaction? We've hit issues with longer running migrations so we generally try to make them idempotent.

lib/strong_migrations/migration_helpers.rb
```diff
@@ -73,16 +75,172 @@ def remove_null_constraint_safely(table_name, column_name, name: nil)
...
+              change_column_default(table_name, column_name, default)
+            end
+
+            default_after_type_cast = connection.type_cast(default)
```
jturkel: How hard would it be to support default expression e.g. `uuid_generate_v4()`? We've needed to do that pretty often as we've migrated away from `bigint` identifiers (at least for what we externally expose).

lib/strong_migrations/migration_helpers.rb
```diff
@@ -73,16 +75,172 @@ def remove_null_constraint_safely(table_name, column_name, name: nil)
...
+        end
+      end
+    end
+
+    def backfill_column_safely(table_name, column_name, value, batch_size: 1000)
+      ensure_not_in_transaction(__method__)
+
+      table = Arel::Table.new(table_name)
+      count_arel = table.project(Arel.star.count.as("count"))
+      total = connection.exec_query(count_arel.to_sql).first["count"]
```
jturkel: We've seen `count(*)` on big tables take a long time. Should we avoid this count query and just handle the `start_arel` returning a null result?

lib/strong_migrations/migration_helpers.rb
```diff
@@ -73,16 +75,172 @@ def remove_null_constraint_safely(table_name, column_name, name: nil)
...
+
+      return if total == 0
+
+      primary_key = connection.primary_key(table_name)
+
+      start_arel = table
```
jturkel: Following up on my previous comment about restartability of migrations, should we skip rows that already have a non-null value for `column_name`? It's not strictly required but can be a really helpful for long running migrations that get restarted due to a network blip, Rails process dying, etc.

lib/strong_migrations/migration_helpers.rb
```diff
@@ -73,16 +75,172 @@ def remove_null_constraint_safely(table_name, column_name, name: nil)
...
+        finish_arel = table
+          .project(table[primary_key])
+          .where(table[primary_key].gteq(start_pk))
+          .order(table[primary_key].asc)
+          .skip(batch_size)
+          .take(1)
+
+        finish_result = connection.exec_query(finish_arel.to_sql).first
+
+        update_arel = Arel::UpdateManager.new
```
jturkel: Our app sometimes updates multiple rows which makes migrations like this deadlock prone. We always force locks to be acquired in primary key order to avoid this. Not sure how common that practice is and whether or not it's worth folding into the gem to avoid deadlocks.

lib/strong_migrations/migration_helpers.rb
fatkodima: @jturkel Thanks for your suggestions. Implemented them in #105

lib/strong_migrations/migration_helpers.rb
fatkodima: Implemented in #105

lib/strong_migrations/migration_helpers.rb
fatkodima: Implemented in #105

lib/strong_migrations/migration_helpers.rb
fatkodima: It is not *that* hard, but it definitely will introduce some ugly code changes in `backfill_column_safely` method. Let's wait for more demand and implement it in separate PR, if needed.

lib/strong_migrations/migration_helpers.rb
fatkodima: Didn't get it. Can you provide examples and elaborate a little more?

ankane:

PR #108, opened 2020-01-01
Change column null safely

fatkodima:
[Real changes](https://github.com/ankane/strong_migrations/commit/efc1985621014d5a3205178e65ea17b928947016) of this pr.


ankane:

PR #109, opened 2020-01-02
Add optional reversible migrations check

fatkodima:
```ruby
StrongMigrations.enable_check(:reversible_migrations)
```


fatkodima:
@ankane Wdyt on this? Any suggestions?

ankane:
Hey @fatkodima, sorry for the long delay. After thinking about it for a while, I don't think it's something I'd like to include in Strong Migrations right now.

fatkodima:
No problems.

PR #110, opened 2020-01-02
Add ability to catch unsafe migrations during rollback

fatkodima:
Related #95

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 0.5.2 (unreleased)
 
+- Add ability to catch unsafe migrations during rollback
 - Added `add_foreign_key_safely` method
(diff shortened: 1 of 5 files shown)


ankane:
This feature is a bit different than how it's implemented in this PR. Checks would still happen on the `up` phase, but it would check for things like:

```ruby
remove_index :users, :some_column
```

which is not safe to revert without `algorithm: :concurrently`.

ankane:
Hey @fatkodima, I resolved the conflict and merged. https://github.com/ankane/strong_migrations/commit/e80bfa4872e31c55e1c0bc3ab5e77528c075b45d

I renamed the option to `check_down` to indicate it runs for all down migrations (not just rollbacks) and added tests for migrations that use the `change` method.

Thanks for adding this!

PR #112, opened 2020-01-22
Update note re changing varchar to text

mlissner:
I confess I'm only about 70% sure of this, but I wanted to propose it anyway in the hopes that somebody here might have thoughts about it. Please consider this PR a suggestion until we're certain, lest my bad information spread.

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,7 +301,7 @@ class ChangeSomeColumnType < ActiveRecord::Migration[6.0]
 end
 ```
 
-One exception is changing a `varchar` column to `text`, which is safe in Postgres.
+One exception is changing a `varchar` column to `text`, which is safe in Postgres, though it rewrites any indexes on the column, which requries an exclusive lock.
 
 #### Good
 


ankane:
Hey @mlissner, can you explain more about why you believe this is the case?

From what I understand, this operation only removes the length constraint.

If you run the following statements, the `ALTER COLUMN` operation takes the same amount of time with and without an index.

```psql
\timing
DROP TABLE IF EXISTS t1;
CREATE TABLE t1 (id serial PRIMARY KEY, name character varying);
INSERT INTO t1 (id) SELECT generate_series(1,1000000) i;
CREATE INDEX ON t1 (name);
ALTER TABLE t1 ALTER COLUMN name TYPE text;
```

mlissner:
OK, this doesn't mention the exclusive lock (I'm pretty sure I got that from StackOverflow, but I've lost the post), but the [Postgresql `ALTER TABLE` docs][1] have this to say (emphasis mine):

> [...] changing the type of an existing column will require the entire table and its indexes to be rewritten. As an exception when changing the type of an existing column, if the USING clause does not change the column contents and the old type is either binary coercible to the new type or an unconstrained domain over the new type, a table rewrite is not needed; but **any indexes on the affected columns must still be rebuilt.**

[1]: https://www.postgresql.org/docs/10/sql-altertable.html

Not sure why that doesn't have a `timing` impact for your test. I thought it might be because you had `varying` without a length constraint, so I tried:

    CREATE TABLE t1 (id serial PRIMARY KEY, name character varying(30));

But that didn't change things either. 

I'm perplexed. Am I missing something obvious here? Docs seem pretty clear.

ankane:
Yeah, the docs definitely seem to indicate indexes must be rebuilt. Either:

1. The test case doesn't trigger it OR
2. The docs are outdated

I made the test more comprehensive, but the `ALTER TABLE` still appears to be too fast to rebuild the index and the index is still used.

```psql
\timing
DROP TABLE IF EXISTS t1;
CREATE TABLE t1 (id serial PRIMARY KEY, name character varying);
INSERT INTO t1 (id, name) SELECT i, 'test' || i FROM generate_series(1,1000000) i;
CREATE INDEX ON t1 (name);
ANALYZE t1;
EXPLAIN ANALYZE SELECT * FROM t1 WHERE name = 'test10000';
ALTER TABLE t1 ALTER COLUMN name TYPE text;
ANALYZE t1;
EXPLAIN ANALYZE SELECT * FROM t1 WHERE name = 'test10000';
```

(fwiw, I used `character varying` since that's the type Rails uses for `string` columns)

You could try asking on the PostgreSQL mailing list if it's still a concern.

mlissner:
I'll try the mailing list. It seems like the only way to untie this knot.

mlissner:
OK, so the docs were wrong according to the conversation here: 

https://www.postgresql.org/message-id/flat/CAMp9%3DExXtH0NeF%2BLTsNrew_oXycAJTNVKbRYnqgoEAT01t%3D67A%40mail.gmail.com

I mean, of course they were! Why would we trust docs over our very eyes, which shows that the index wasn't being rewritten! 

I've filed a patch to postgresql's hacker email list. We'll see where it goes.

Thanks for the help developing a test for this.

ankane:
No problem, thanks for getting to the bottom of it. For reference, here's the awesome commit mentioned in the mailing list that made this possible (for Postgres 9.2): https://github.com/postgres/postgres/commit/367bc426a

mlissner:
What blows my mind is how it went unfixed for eight years. Perhaps they have better testing infrastructure than we do, but whether an index rewrite is needed is no small thing for us at least.

ankane:
Incorrect docs happen. From my experience, varchar to text conversion has just worked, so there hasn't been a need to check the docs - maybe it was the same for others.

PR #115, opened 2020-01-28
Set Postgres statement and lock timeouts in milliseconds

geoffharcourt:
Postgres expects a number argument for statement and lock timeouts to be
in milliseconds. This results in the recently released migration-specific
timeouts being shorter than intended if the user sets them with
`StrongMigrations.statement_timeout=` and `.lock_timeout=`, as PG
interprets seconds as milliseconds.

This change updates the checker to multiply the submitted value by 1,000 
if the database is PG.

PG docs:
https://www.postgresql.org/docs/11/runtime-config-client.html

Co-authored-by: Susanna Kosonen <kosonen.susanna@gmail.com>

### DIFF
--- a/test/strong_migrations_test.rb
+++ b/test/strong_migrations_test.rb
@@ -273,7 +273,7 @@ class CheckTimeouts < TestMigration
 
     $statement_timeout =
       if postgresql?
-        connection.select_all("SHOW statement_timeout").first["statement_timeout"].to_i
+        connection.select_all("SHOW statement_timeout").first["statement_timeout"]
       elsif mysql?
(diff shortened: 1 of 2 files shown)


test/strong_migrations_test.rb
```diff
@@ -273,7 +273,7 @@ def change
 
     $statement_timeout =
       if postgresql?
-        connection.select_all("SHOW statement_timeout").first["statement_timeout"].to_i
+        connection.select_all("SHOW statement_timeout").first["statement_timeout"]
```
geoffharcourt: The results here get set as "10s" and "1h", so we can't call `#to_i`.

ankane:
Nice catch and great code! Thanks @geoffharcourt and Susanna. Will push out a new release shortly.

PR #120, opened 2020-05-17
Experimental support for SQL migration inspection on Postgres

joehorsnell:
Opening for feedback on the general idea.

This PR is a proof-of-concept for analysing the safety of raw SQL migrations
using [pg_query][1]. In this case, showing the safety of a `NOT VALID` foreign
key addition vs the unsafe option.

Thoughts? One obvious downside is that this is Postgres-specific, so MySQL users
would be out of luck. However, that probably shouldn't be a reason on its own to
not add support for Postgres. I think this could be adapted even to analyse the
SQL generated by the Rails migrations DSL methods too?

[1]: https://github.com/lfittl/pg_query

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+- Added experimental support for safety checking raw SQL migrations on Postgres with the `inspect_sql_postgresql` option
+
 ## 0.6.8 (2020-05-13)
 
 - `change_column_null` on a column with a `NOT NULL` constraint is safe in Postgres 12+
--- a/Gemfile
+++ b/Gemfile
@@ -3,3 +3,9 @@ source "https://rubygems.org"
 gemspec
 
 gem "activerecord", "~> 6.0.0"
+
+group :development, :test do
+  gem "pg_query", "~> 1.2"
+  gem "pry-byebug"
+  gem "awesome_print"
+end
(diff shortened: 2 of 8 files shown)


ankane:
Hey @joehorsnell, thanks for the PR! This is a neat idea, but since most migrations won't use SQL directly, I don't think it's worth the extra complexity. That being said, you could add it as a custom check:

```ruby
StrongMigrations.disable_check(:execute)
  # code
end
```

joehorsnell:
Hey @ankane, thanks for the quick feedback, appreciated 👍.

> I think this could be adapted even to analyse the SQL generated by the Rails migrations DSL methods too?

Not sure if you saw this? I was thinking that perhaps the same mechanism, of analysing the SQL directly, could potentially be used for *all* migrations, ie not just “execute” ones, either to complement, or even maybe replace some of the current checks?

This would obviously require intercepting and inspecting the SQL generated by the Rails migration DSL methods before it runs.

Also obviously, this would be quite a departure from how things work currently in Strong Migrations, so I can understand if it’s not a direction you want to go in.

Overall this approach I think would be more complex and therefore difficult to implement, but has the potential to be more accurate and certainly more general, as it would work no matter how the SQL was ultimately generated.

While much simpler, I guess there are limitations with the current approach of looking for Rails DSL method calls, eg if any custom methods are used, or in my example the use of execute.

Let me know what you think. Happy to close if this is not a direction you want to go in. I might look at adding this functionality to a separate gem, which could potentially hook into SM. Maybe you could link to? Thanks!

ankane:
Thanks for the follow up, but I don't think it's the best direction for Strong Migrations right now.

PR #123, opened 2020-06-02
improvements to documentation for NOT NULL constraints

### LINKED ISSUE #122: Making an existing column NOT NULL in PG 12+
There's a hidden gem in PG 12 that allows making an existing column NOT NULL without a table scan.

I have a recipe here: https://gist.github.com/jjb/fab5cc5f0e1b23af28694db4fc01c55a

There are links to explanation/discussion at the bottom.

This might be worth mentioning in the readme?

jjb:
closes #122

### DIFF
--- a/README.md
+++ b/README.md
@@ -479,7 +479,7 @@ end
 
 #### Good
 
-Instead, add a constraint:
+Instead, add a constraint (this operation locks, but because of `NOT VALID` it will not do a table scan):
 
 ```ruby
 class SetSomeColumnNotNull < ActiveRecord::Migration[6.0]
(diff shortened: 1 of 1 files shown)


ankane:
Hey @jjb, thanks for the PR. Can you explain more about what this tries to accomplish? All schema operations create locks, and I don't think there's any benefit to moving it to a separate migration.

jjb:
I was trying to make it more clear that everything up to the last migration is beneficial and suitable for PG < 12, and then the last one is ideal for 12+. I think maybe the comment currently in master is a bit too subtle. Instead of a separate migration, the comments could be expanded a bit.

ankane:
Tried to rework so it's a little less subtle (but still concise): https://github.com/ankane/strong_migrations/commit/81481baac9751ad252d02f77b962558a00f04b3a

jjb:
Looks great, thanks!

PR #127, opened 2020-07-06
Add reference with index and reference

maschwenk:
Suggesting disabling transactions when adding a reference seems wrong. We'd never want to disable transactions around adding a column so why would we have special logic for a reference?


ankane:
Hey @maschwenk, transactions must be disabled to add an index concurrently. You could break it into two separate migrations (add column and add index) and only disable transactions for the 2nd, but I don't think it's needed. However, you can add a custom check if it's something you'd prefer for your application.

PR #132, opened 2020-10-12
Link to example of how to "write to both columns"

ghiculescu:
I found myself having a bit of a mind blank on how to do this. Adding a link to an [SO answer](https://stackoverflow.com/questions/63295941/how-can-i-realise-write-to-both-columns-in-rails-active-record-as-strong-migr/63304987#63304987) that steps through it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,7 +246,7 @@ end
 A safer approach is to:
 
 1. Create a new column
-2. Write to both columns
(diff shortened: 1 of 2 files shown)


ankane:
Hey @ghiculescu, thanks for the PR (and sorry for the long delay), but it's not something I'd like to include.

PR #141, opened 2021-03-02
update readme to only install in development group

Zhong-z:
>Catch unsafe migrations in development

Seems no need to include this in production.


Zhong-z:
closing, just saw this https://github.com/ankane/strong_migrations/pull/68

PR #144, opened 2021-03-26
Add clarity to the 'Removing a column' section

karimtimer:
I wasn't sure whether to do this before or after step 4. Adding this to help anyone else who has a similar mind blank!


ankane:
Hey @karimtimer, sorry for the delay. Looks great!

PR #149, opened 2021-05-23
Support multiple databases for `target_version`

fatkodima:
It is very WIP. It currently only allows to set different database versions via

```ruby
StrongMigrations.target_version = { primary: 10, animals: 11 }
```

Considerations:
* Proper testing needs to be added.
* I think, we should also allow other existing configurations to be extended to support multiple databases. There are 2 approaches I'm thinking about:
1) via hashes
```ruby
StrongMigrations.target_version = {primary: 10, animals: 11}
StrongMigrations.start_after = {primary: 20170101000000, animals: 20170101000123}
# ... other configs
``` 
2) via methods calling
```ruby
StrongMigrations.primary.target_version = 10
StrongMigrations.animals.target_version = 11

StrongMigrations.primary.start_after = 20170101000000
StrongMigrations.animals.start_after = 20170101000123

# ... other configs
```
* I haven't found a better way to detect specification names ("primary", "animals") within migrations than manually setting it via ENV within a custom rake task and make existing rails tasks to depend on it.

### DIFF
--- a/lib/strong_migrations/adapters/abstract_adapter.rb
+++ b/lib/strong_migrations/adapters/abstract_adapter.rb
@@ -50,7 +50,19 @@ module StrongMigrations
         target_version ||= StrongMigrations.target_version
         version =
           if target_version && StrongMigrations.developer_env?
-            target_version.to_s
+            if target_version.is_a?(Hash)
+              if ActiveRecord::VERSION::STRING.to_f < 6.1
+                raise StrongMigrations::Error, "StrongMigrations.target_version does not support multiple databases for ActiveRecord < 6.1"
+              end
+
+              target_version = target_version.stringify_keys
+              db_config_name = connection.pool.db_config.name
+              target_version.fetch(db_config_name) do
+                raise StrongMigrations::Error, "StrongMigrations.target_version is not configured for :#{db_config_name}"
+              end
+            else
+              target_version.to_s
+            end
           else
             yield
           end
--- a/test/misc_test.rb
+++ b/test/misc_test.rb
@@ -47,4 +47,83 @@ class MiscTest < Minitest::Test
       end
     end
   end
+
+  def test_target_version_multiple_dbs_below_6_1
+    skip if multiple_dbs?
+
+    with_target_version({ primary: 10, animals: 15.0 }) do
+      error = assert_raises(StrongMigrations::Error) do
+        assert_safe AddColumnDefault
+      end
+      assert_equal "StrongMigrations.target_version does not support multiple databases for ActiveRecord < 6.1", error.message
+    end
+  end
+
+  def test_target_version_multiple_dbs
+    skip unless multiple_dbs?
+
+    with_multiple_dbs do
+
+      safe_version = postgresql? ? 11 : (mysql? ? "8.0.12" : "10.3.2")
+      with_target_version({ primary: safe_version }) do
+        assert_safe AddColumnDefault
+      end
+
+      unsafe_version = postgresql? ? 10 : (mysql? ? "8.0.11" : "10.3.1")
+      with_target_version({ primary: unsafe_version }) do
+        assert_unsafe AddColumnDefault
+      end
+    end
+  end
+
+  def test_target_version_multiple_dbs_unconfigured
+    skip unless multiple_dbs?
+
+    with_multiple_dbs do
+      error = assert_raises(StrongMigrations::Error) do
+        with_target_version({ animals: 10 }) do
+          assert_safe AddColumnDefault
+        end
+      end
+      assert_equal "StrongMigrations.target_version is not configured for :primary", error.message
+    end
+  end
+
+  private
+
+  def with_multiple_dbs(&block)
+    previous_db_config =
+      if ar_version >= 6.1
+        ActiveRecord::Base.connection_db_config.configuration_hash
+      else
+        ActiveRecord::Base.connection_config
+      end
+
+    multi_db_config = {
+      "test" => {
+        "primary" => {
+          "adapter" => $adapter,
+          "database" => "strong_migrations_test"
+        },
+        "animals" => {
+          "adapter" => $adapter,
+          "database" => "animals_test"
+        }
+      }
+    }
+    ActiveRecord::Base.configurations = multi_db_config
+
+    ActiveRecord::Base.connects_to(database: { writing: :primary })
+    ActiveRecord::Base.connected_to(role: :writing, &block)
+  ensure
+    ActiveRecord::Base.establish_connection(previous_db_config)
+  end
+
+  def multiple_dbs?
+    ar_version >= 6.1
+  end
+
+  def ar_version
+    ActiveRecord::VERSION::STRING.to_f
+  end
 end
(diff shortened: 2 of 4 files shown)


fatkodima:
@ankane I am going to finish this PR.

> I haven't found a better way to detect specification names ("primary", "animals") within migrations than manually setting it via ENV within a custom rake task and make existing rails tasks to depend on it.

So this is not a problem currently, I already found how to overcome this.

The API will look like:
```ruby
StrongMigrations.target_version = {primary: 10, animals: 11}
StrongMigrations.start_after = {primary: 20170101000000, animals: 20170101000123}
```

I think just this 2 options will cover 95% of the per-database configuration people will need, so not sure more options are needed.

Wdyt?

ankane:
Sounds good to me. Let's just do `target_version` for now since I think that's the main one that'll be different between DBs.

fatkodima:
@ankane Ready for review. I also did a small cleanup (`Utils` module).

ankane:
Thanks @fatkodima! The current changeset looks pretty large. It'd be good to limit the `lib` changes to `AbstractAdapter#target_version` and the test changes to `test/misc_test.rb`.

fatkodima:
Should I extract `support/database.yml` related changes and `Utils` related changes as separate PRs? 
Both were introduced to make the changes relevant to the feature cleaner.

ankane:
I'm not really sure either are needed.

fatkodima:
Updated.

lib/strong_migrations/adapters/abstract_adapter.rb
```diff
@@ -47,7 +47,15 @@ def select_all(statement)
...
+            StrongMigrations.target_version.fetch(db_config_name) do
+              raise "StrongMigrations.target_version is not configured for :#{db_config_name}"
+            end
+          else
+            StrongMigrations.target_version
+          end
+
         version =
           if target_version && StrongMigrations.developer_env?
             target_version.to_s
```
ankane, line 61: I think it's better to put all the new logic here so it's not executed unless it'll be used.

lib/strong_migrations.rb
```diff
@@ -27,8 +27,9 @@ class UnsupportedVersion < Error; end
   class << self
     attr_accessor :auto_analyze, :start_after, :checks, :error_messages,
       :target_postgresql_version, :target_mysql_version, :target_mariadb_version,
-      :enabled_checks, :lock_timeout, :statement_timeout, :check_down, :target_version,
+      :enabled_checks, :lock_timeout, :statement_timeout, :check_down,
```
ankane, line 30: Please don't make any changes to this file.

test/test_helper.rb
```diff
@@ -12,14 +12,25 @@ def self.env
 end
 
 $adapter = ENV["ADAPTER"] || "postgresql"
-connection_options = {
-  adapter: $adapter,
-  database: "strong_migrations_test"
+
```
ankane, line 15: Please don't make any changes to this file.

ankane (commented):
Thanks @fatkodima, it's looking better. Added a few comments inline.

test/test_helper.rb
fatkodima: We need to restore the connection back after the test - https://github.com/ankane/strong_migrations/pull/149/files#diff-c8ef59150a5fd973aeaded97986349c50366f7184bc49b7e847d43420496c146R110
How should I do that without the changes in this file?

ankane: You can get the config with `ActiveRecord::Base.connection_db_config.configuration_hash` for Active Record 6.1+ and `ActiveRecord::Base.connection_config` for < 6.1.

Also, it looks like some of the tests are skipped for MySQL and MariaDB. `target_version` is supported for both of them, so it'd be good to include them (like the existing tests that use `target_version`).

fatkodima:
@ankane Updated with proposed suggestions. Please, take a look.

lib/strong_migrations/adapters/abstract_adapter.rb
```diff
@@ -50,12 +50,32 @@ def target_version(target_version)
...
           if target_version && StrongMigrations.developer_env?
-            target_version.to_s
+            if target_version.is_a?(Hash)
+              if ActiveRecord::VERSION::STRING.to_f < 6.0
+                raise "Multiple databases are not supported by this ActiveRecord version"
+              end
+
+              target_version = target_version.stringify_keys
+              target_version.fetch(db_config_name) do
+                raise "StrongMigrations.target_version is not configured for :#{db_config_name}"
```
ankane, line 60: I think it'd be better to raise a `StrongMigrations::Error` instead of a `RuntimeError`

lib/strong_migrations/adapters/abstract_adapter.rb
```diff
@@ -50,12 +50,32 @@ def target_version(target_version)
...
           else
             yield
           end
+
         Gem::Version.new(version)
       end
+
+      def db_config_name
+        if ActiveRecord::VERSION::STRING.to_f < 6.1
+          connection.pool.spec.name
```
ankane, line 74: This always returns `primary` when running migrations in Rails 6 apps, so this feature may need to be limited to Rails 6.1+

test/misc_test.rb
```diff
@@ -47,4 +47,83 @@ def test_target_version_outside_developer_env
...
+          assert_safe AddColumnDefault
+        end
+      end
+      assert_equal "StrongMigrations.target_version is not configured for :primary", error.message
+    end
+  end
+
+  private
+
+  def with_multiple_dbs(connects_to: :primary, &block)
```
ankane, line 94: `connects_to` is always `:primary` for the current tests, so no need for a keyword argument.

lib/strong_migrations/adapters/abstract_adapter.rb
```diff
@@ -50,12 +50,32 @@ def target_version(target_version)
...
+              target_version.fetch(db_config_name) do
+                raise "StrongMigrations.target_version is not configured for :#{db_config_name}"
+              end
+            else
+              target_version.to_s
+            end
           else
             yield
           end
+
```
ankane, line 68: Please remove / keep the changeset minimal

ankane (commented):
Thanks @fatkodima, looks much better. Added some more comments inline.

lib/strong_migrations/adapters/abstract_adapter.rb
fatkodima: Can you clarify (I am not sure I got it)? 
This if-else is used because there was an API change some time ago - https://github.com/rails/rails/pull/37253.

fatkodima:
@ankane Updated. 
We spent quite some time back and forth in this PR, so in future if you feel like it is easier/faster to just push changes to this PR than spending time on review comments, feel free to do so.

lib/strong_migrations/adapters/abstract_adapter.rb
ankane: If you create a new Rails 6.0 app, set up multiple databases, and call `connection.pool.spec.name` inside a migration, it always returns `primary` when you run `rails db:migrate` (I'm guessing this is due to how Rails 6.0 internally handles connection switching between migrations).

fatkodima: Yes, indeed. Thank you. Pushed new changes.

ankane:
Looks great, thanks!

PR #161, opened 2021-09-23
Add enabled_tabels option

gogochikago:
I would like to have functionality to restrict checks only for special tables.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.7.9 (2021-09-22)
+
+- Added `enabled_tables` option
+
 ## 0.7.8 (2021-08-03)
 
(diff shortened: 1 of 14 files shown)


ankane:
Hey @gogochikago, thanks for the PR! Can you share more about the use case?

gogochikago:
@ankane Hello! Thanks for reply
Usually projects have small tables that do not need some checks, for example, updating the default value, etc. I would like to be able to select tables on which to include checks

gogochikago:
Hey @ankane, what do you think about it?

ankane:
Hey @gogochikago, sorry for the delay. There are some operations (like removing a column or renaming a table/column) that I think we'd want to keep regardless of table size. Overall, it's a neat idea, but I don't think it's worth the additional complexity.

PR #165, opened 2021-11-01
[idea] allow for enforcing reason for overriding with safety_assured

okhwaja:
we're interested in requiring all migrations that override safety checks to explain why it is safe to do so ("this table is unused so a full table rewrite is fine", etc.). Having the rationale directly in the code helps reviewers and future readers understand why it was safe at that time (as opposed to an offline conversation)

This PR adds an optional `reason` arg to `safety_assured` and enforces that it must be provided if the `require_safety_reason` option is enabled. it's disabled by default for backwards compatibility

curious what you think. thanks!

### DIFF
--- a/lib/strong_migrations.rb
+++ b/lib/strong_migrations.rb
@@ -19,13 +19,14 @@ module StrongMigrations
     attr_accessor :auto_analyze, :start_after, :checks, :error_messages,
       :target_postgresql_version, :target_mysql_version, :target_mariadb_version,
       :enabled_checks, :lock_timeout, :statement_timeout, :check_down, :target_version,
-      :safe_by_default
+      :safe_by_default, :require_safety_reason
     attr_writer :lock_timeout_limit
   end
   self.auto_analyze = false
   self.start_after = 0
   self.checks = []
   self.safe_by_default = false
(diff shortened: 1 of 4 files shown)


mmontagna (approved):

levkk:
I would even go as far as to make it mandatory.

ankane:
Hey @okhwaja, thanks for the PR! It's an interesting idea. Some operations like removing a column or setting not null on an existing column before Rails 6.1 require `safety_assured` to be done safely, so it probably shouldn't be required for those operations. Also, it'd be good to run this with the team for a few months to understand how people use it before deciding to support it.

sherin:
Hi @ankane We had several issues recently where risky migrations have managed to get through because engineers would just copy older migrations (along with the safety_assured wrap). There has also been some confusion that adding safety_assured itself assures safety :) Both of these are understandable and expose loopholes in the current design.

Any thoughts on how to improve the effectiveness of strong migrations? I think Osman's way of adding some friction is great but may be there is something better.

ankane:
Thanks for the additional context. I think trying the approach internally makes sense (but want to understand how it goes before making a decision to merge). It sounds like renaming the method could also be something to try.

PR #166, opened 2021-11-03
update database_tasks for active record 7

CharlesDelannoy:
This PR concerns only rails app with minimum version 7, or said other way with ActiveRecord 7 minimum. 

This aims at fixing the multi db handling with the 'new' migrate method that can now have one argument => `version`
👉  https://github.com/rails/rails/blob/main/activerecord/lib/active_record/tasks/database_tasks.rb#L272

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,6 @@ gemspec
 
 gem "rake"
 gem "minitest", ">= 5"
-gem "activerecord", "~> 6.1.0"
+gem "activerecord", "~> 7.0.0.alpha2"
 gem "pg"
 gem "mysql2"
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,6 @@ gemspec
 
 gem "rake"
 gem "minitest", ">= 5"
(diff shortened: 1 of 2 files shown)


ankane:
Hey @CharlesDelannoy, thanks for the PR! I think changing line 3 to `def migrate(*)` and leaving line 4 as is should keep it backward compatible.

ankane:
Also, just added a Gemfile for AR 7 in e36dace221ba975c4bfb994fcdbf597df193fef1.

CharlesDelannoy:
Hi @ankane, I made the changes according to your comments. Thanks!

ankane:
Thanks @CharlesDelannoy. Line 4 shouldn't need any changes since `super` without parentheses will forward arguments.

CharlesDelannoy:
@ankane your're right, it's done !

ankane:
Great, thanks @CharlesDelannoy!

EQuincerot:
@ankane do you think you could release this soon?
With Charles we already use rails 7, and need this change to avoid a monkey patch.

PR #175, opened 2022-02-16
Explain NOT NULL DDL on MySQL.

ryotarai:
MySQL 5.6 or later supports online DDL and it does not lock a table. This PR explains it in README.
(Please point out if my English is wrong or unnatural)

> To avoid accidentally making the table unavailable for reads, writes, or both, specify a clause on the [ALTER TABLE](https://dev.mysql.com/doc/refman/5.6/en/alter-table.html) statement such as LOCK=NONE (permit reads and writes) or LOCK=SHARED (permit reads).
(https://dev.mysql.com/doc/refman/5.6/en/innodb-online-ddl.html)

### DIFF
--- a/README.md
+++ b/README.md
@@ -431,7 +431,17 @@ end
 
 #### Good - MySQL and MariaDB
 
-[Let us know](https://github.com/ankane/strong_migrations/issues/new) if you have a safe way to do this.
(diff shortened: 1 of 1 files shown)


ankane:
Hey @ryotarai, thanks for sharing and the PR! This is super helpful.

From my testing and interpretation of the docs, I think MySQL may default to `ALGORITHM=INPLACE, LOCK=NONE` as long as `sql_mode` is `STRICT_ALL_TABLES` or `STRICT_TRANS_TABLES`. Has that been your experience?

(I'm thinking we may be able to allow `change_column_null` as long as one of those is set if that's the case. It's not as ideal as explicitly specifying it, but would be more convenient for developers).

ryotarai:
> From my testing and interpretation of the docs, I think MySQL may default to `ALGORITHM=INPLACE, LOCK=NONE` as long as `sql_mode` is `STRICT_ALL_TABLES` or `STRICT_TRANS_TABLES`.

Yes, I think so too.

> I'm thinking we may be able to allow change_column_null as long as one of those is set if that's the case

I understand. I'll open another PR to implement this.

PR #176, opened 2022-02-16
Allow setting NOT NULL on MySQL with online DDL.

ryotarai:
Discussion: https://github.com/ankane/strong_migrations/pull/175

This PR allows setting NOT NULL on MySQL.
If any of the following conditions meets, migration will be aborted:
* MySQL is older then 5.6 (Online DDL is supported on MySQL 5.6 or later)
* MariaDB is older then 10.0 (Online DDL is supported on MariaDB 5.6 or later)
* strict mode is not enabled
* default value is set

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,7 +358,7 @@ end
 
 :turtle: Safe by default available
 
-#### Bad
+#### Bad - Postgres
 
 Setting `NOT NULL` on an existing column blocks reads and writes while every row is checked.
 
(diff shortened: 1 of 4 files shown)


ankane:
Looks really great, thanks @ryotarai!

For the option, I think we can use `target_sql_mode` for a slightly simpler name (which takes a string instead of an array), so I'll make that change in a follow-up commit.

Also, MySQL < 5.7 and MariaDB < 10.1 aren't supported, but this wasn't really clear before https://github.com/ankane/strong_migrations/commit/b67db98170ca3b57622eb125be09beb68c43956e, so I think it makes sense to keep the version checks until 0.9.0.

PR #189, opened 2022-07-08
Check for PostgreSQL index corruption in all Rails environments

jdufresne:
Previously, the PostgreSQL index corruption bug was checked only in
production environments.

However, this can cause the issue to go unnoticed up to the point where
a project tries to release to production and migrate the database. CI
systems often run in the Rails test environment causing this check to
silently be skipped. This scenario happened to a project of mine,
requiring quick action to undo a series of steps we thought were being
checked in CI.

Projects can continue to use the StrongMigrations.target_version setting
if their developer or test database version differs from production.

### DIFF
--- a/lib/strong_migrations/adapters/postgresql_adapter.rb
+++ b/lib/strong_migrations/adapters/postgresql_adapter.rb
@@ -155,11 +155,9 @@ module StrongMigrations
         select_all(query.squish).any?
       end
 
-      # only check in non-developer environments (where actual server version is used)
       def index_corruption?
         server_version >= Gem::Version.new("14.0") &&
-          server_version < Gem::Version.new("14.4") &&
-          !StrongMigrations.developer_env?
+          server_version < Gem::Version.new("14.4")
       end
 
       private
--- a/test/add_index_test.rb
+++ b/test/add_index_test.rb
@@ -84,10 +84,8 @@ class AddIndexTest < Minitest::Test
 
   def test_corruption
     skip unless postgresql?
-    outside_developer_env do
-      with_target_version(14.3) do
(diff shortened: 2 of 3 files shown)


jdufresne:
Hi @ankane Thanks for the response.

Would you be willing to explain the design decision a bit? Had this check been more universal, my project would have caught the index corruption issue earlier during CI. Instead, the issue wasn't seen until after a database upgrade. Luckily, we were able to revert the upgrade without an issue, but I think it would have been more convenient for my project (and I'm hoping others) if this kind of thing had been louder.

WDYT about projects can't using `StrongMigrations.target_version` that have different versions in production, testing, and development?

ankane:
It's designed this way to minimize effort for most users. If the check affected non-developer environments:

1. Without `target_version`, developers would need to upgrade their development and test databases on affected versions. 
2. With `target_version`, apps on Postgres 14 (now and in the future) would need to update `target_version` to include the minor version.

I don't think the issue described above is enough to warrant the additional effort. Ideally, the only effort needed for this is to update production/staging databases once.

jdufresne:
> Without target_version, developers would need to upgrade their development and test databases on affected versions.

Tools like Docker make this relatively easy. I understand that it is more effort than doing nothing but hopefully still low effort.

I also think this would would be a good thing for the health of a project. If a project depends on a particular version of PostgreSQL in production then the project should be testing against that same version in testing and development. To do otherwise risks unexpected compatibility issues on release (or in this case, unexpected corruption bugs).

> With target_version, apps on Postgres 14 (now and in the future) would need to update target_version to include the minor version.

Yeah, I agree, but that is true of using `target_version` in general. My projects don't use `target_version` for this reason, but I imagine anyone using it would need to understand this maintenance issue.

PR #195, opened 2022-08-19
Check for string column function defaults

duderman:
Recently we've had an issue with adding a uuid column with a default to fill it up during migration. Which resulted in DB being locked during migration because apparently if you use a function for a default value of a column it evaluates this function during the migration for each existing row. This check can help to prevent locking the DB in such cases

### DIFF
--- a/lib/strong_migrations/checker.rb
+++ b/lib/strong_migrations/checker.rb
@@ -71,6 +71,8 @@ module StrongMigrations
           # and no longer eligible to be retried at migration level
           # okay to have false positives
           @committed = true
+        when :change_column_default
+          check_change_column_default(*args)
         end
 
         # custom checks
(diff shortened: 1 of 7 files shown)


fatkodima:
`change_column_default` is free for PostgreSQL, because it updates only the internal metadata.

But `add_column` with default is not. For pg < 11 it is always unsafe to add a column with a default. For pg >= 11 it is unsafe when the default is "volatile". 

I think, the check for volatility in `add_column` should be improved. You can look at an example on how to do this here - https://github.com/fatkodima/online_migrations/blob/c38e5a2976da45baf61180dac97b6664949b99e6/lib/online_migrations/command_checker.rb#L177-L178

ankane:
Hey @duderman, thanks for the PR. As @fatkodima mentioned, `change_column_default` is safe since it doesn't update existing rows.

For `add_column`, it looks like there's a [special case in Active Record](https://github.com/rails/rails/blob/50cbcc82e5eef6ba9f070ec2df4561062ea9bb62/activerecord/lib/active_record/connection_adapters/postgresql/quoting.rb#L92-L93) for `uuid` columns that allows functions to be passed directly to `default` (unlike other data types), so think we just need to check for that specific case.

ankane:
Just pushed a check for it. Thanks again for reporting @duderman!

Issue #199, opened 2022-09-21
Weird postgres behaviour with NULL as default value


fabn:
Hi, we're using `strong_migrations` gem and it's very useful, so thanks for your work.

We were just bitten by a small downtime not detected by this gem, it was obviously our fault but it could be added to the gem to save other users.

We deployed this migration in production and it locked our system for about 40 seconds while running.

```ruby
add_column :products, :test_price, :decimal, precision: 16, scale: 4, default: nil
# This will obviously works
add_column :products, :test_price, :decimal, precision: 16, scale: 4
```

We debugged what happens under the hood in postgres (version 9.6) and we found out that these two queries have a different order of magnitude

```
D, [2022-09-21T18:13:32.275103 #49796] DEBUG -- :    (1575.6ms)  ALTER TABLE "products" ADD "test_price_cents" decimal(16,4) DEFAULT NULL
   -> 1.5773s
D, [2022-09-21T18:13:52.684874 #49848] DEBUG -- :    (1.0ms)  ALTER TABLE "products" ADD "test_price_cents" decimal(16,4)
   -> 0.0021s
```

Test were conducted on a development machine with about ~700k records and no other queries running.

So it seems that postgres will actually rewrite whole table even if default value is `NULL`.

I suggest to change the check in [this line](https://github.com/ankane/strong_migrations/blob/b978c08d9e68dd8de9042abb2c6e573822a9f740/lib/strong_migrations/checks.rb#L37) to something like `options.has_key?(:default)` instead of `default.nil?` to prevent this issue for future users.

Thanks.

fabn:
BTW same query in postgres 14 leads to these times

```
D, [2022-09-21T18:31:31.603984 #51664] DEBUG -- :    (1.0ms)  ALTER TABLE "products" ADD "test_price_cents" decimal(16,4) DEFAULT NULL
   -> 0.0016s
D, [2022-09-21T18:30:10.932969 #51525] DEBUG -- :    (0.8ms)  ALTER TABLE "products" ADD "test_price_cents" decimal(16,4)
   -> 0.0015s
```

ankane:
Hey @fabn, thanks for reporting! Just pushed a fix. This affects Postgres < 11. The current version of Strong Migrations only supports Postgres 10+, so you'll probably want to fork the gem and apply a similar fix if your application uses 9.6.

PR #200, opened 2022-10-06
Improve error message on `add_column ..., default: nil`

pocke:
# Problem

This gem raises an error for an explicit null default value since https://github.com/ankane/strong_migrations/issues/199.
The error message is the same as the non-null's one. But there are differences between null and non-null messages.

* The message for non-null says, `Adding a column with a **non-null** default blocks writes while the entire table is rewritten`. It is not appropriate for null default.
* The message displays `change_column_default :table, :column, nil`. It is unnecessary for null default.
* This message displays how to backfill. It is unnecessary too.

# Solution

Update the error message for null default. There are many differences, so I defined a new key for the message.

# Testing

I'm not sure how to test this change. I couldn't find any tests to assert error messages, so I didn't add a test case for this change.

By the way, I just confirmed the behavior with a test script, which is based on https://github.com/rails/rails/blob/de771c979eca99e37e975a58618dd254fc472986/guides/bug_report_templates/active_record_migrations_gem.rb.

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  # Activate the gem you are reporting the issue against.
  gem "activerecord", "~> 7.0.0"
  gem 'strong_migrations', path: '.'
  gem "sqlite3"
end

require "active_record"
require "minitest/autorun"

# This connection will do for database-independent bug reports.
ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")

ActiveRecord::Schema.define do
  create_table :payments, force: true do |t|
    t.decimal :amount, precision: 10, scale: 0, default: 0, null: false
  end
end

class Payment < ActiveRecord::Base
end

class AddTitleToPayment < ActiveRecord::Migration[7.0]
  def change
    add_column :payments, :title, :string, default: nil
  end
end

class BugTest < Minitest::Test
  def test_migration
    AddTitleToPayment.migrate(:up)
  end
end
```

### before (c8091216e494560d7289ab0c76b330b646923565)

```
  1) Error:
BugTest#test_migration:

Adding a column with a non-null default blocks reads and writes while the entire table is rewritten.
Instead, add the column without a default value, then change the default.

class AddTitleToPayment < ActiveRecord::Migration[7.0]
  def up
    add_column :payments, :title, :string
    change_column_default :payments, :title, nil
  end

  def down
    remove_column :payments, :title
  end
end

Then backfill the existing rows in the Rails console or a separate migration with disable_ddl_transaction!.

class BackfillAddTitleToPayment < ActiveRecord::Migration[7.0]
  disable_ddl_transaction!

  def up
    Payment.unscoped.in_batches do |relation| 
      relation.update_all title: nil
      sleep(0.01)
    end
  end
end
```

### after

```
  1) Error:
BugTest#test_migration:

Adding a column with a null default blocks reads and writes while the entire table is rewritten.
Instead, add the column without a default value, then change the default.

class AddTitleToPayment < ActiveRecord::Migration[7.0]
  def change
    add_column :payments, :title, :string
  end
end
```

### DIFF
--- a/lib/strong_migrations/checks.rb
+++ b/lib/strong_migrations/checks.rb
@@ -44,14 +44,21 @@ module StrongMigrations
 Then add the NOT NULL constraint in separate migrations."
         end
 
-        raise_error :add_column_default,
-          add_command: command_str("add_column", [table, column, type, options.except(:default)]),
-          change_command: command_str("change_column_default", [table, column, default]),
-          remove_command: command_str("remove_column", [table, column]),
-          code: backfill_code(table, column, default, volatile),
-          append: append,
-          rewrite_blocks: adapter.rewrite_blocks,
-          default_type: (volatile ? "volatile" : "non-null")
+        if default.nil?
+          raise_error :add_column_default_null,
+            command: command_str("add_column", [table, column, type, options.except(:default)]),
+            append: append,
+            rewrite_blocks: adapter.rewrite_blocks
+        else
+          raise_error :add_column_default,
+            add_command: command_str("add_column", [table, column, type, options.except(:default)]),
+            change_command: command_str("change_column_default", [table, column, default]),
+            remove_command: command_str("remove_column", [table, column]),
+            code: backfill_code(table, column, default, volatile),
+            append: append,
+            rewrite_blocks: adapter.rewrite_blocks,
+            default_type: (volatile ? "volatile" : "non-null")
+        end
       elsif default.is_a?(Proc) && postgresql?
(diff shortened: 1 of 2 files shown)


ankane:
Looks great, thanks @pocke! I updated the test and changed the error message slightly in 2cb942e635e4e60a3740ed9ab5ccee7dcf2a6c55.

PR #207, opened 2022-11-11
Save other people from embarrassing themselves

jaredbeck:


ankane:
Thanks @jaredbeck, but I don't think it's needed.

PR #210, opened 2022-12-07
Fix instructions for column type change

pyromaniac:
Hey, thanks for a great gem!

I noticed that column type change are exactly the same as for column renaming.


ankane:
Hey @pyromaniac, thanks for the PR. I'm not sure there's always a safe way to update it in-place since Rails caches column types, so the instructions use a new column.

pyromaniac:
Hey @ankane, did you even read the new algorithm? Cause it uses a new column and solves the issue with Rails column cache.

PR #216, opened 2023-02-16
Note that increasing column limit is safe only in MariaDB 10.5+

janko:
At work we're currently running MariaDB 10.4, which unfortunately takes a write lock while it's increasing the column limit. Starting from MariaDB 10.5, the schema change is instantaneous regardless of the table size. I verified that *reducing* the limit still take very long (as it probably does a table rewrite), as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,7 +220,7 @@ Type | Safe Changes
 `time` | Increasing or removing `:precision`
 `timestamptz` | Increasing or removing `:limit`, changing to `datetime` when session time zone is UTC in Postgres 12+
 
-And some in MySQL and MariaDB:
+And some in MySQL and MariaDB 10.5+:
 
 Type | Safe Changes
 --- | ---


ankane:
Hey @janko, thanks for the PR. However, I'm not able to reproduce. Test script (mirrors statements from a fresh Rails app):

```ruby
require "benchmark"
require "mysql2"

client = Mysql2::Client.new(host: "127.0.0.1", username: "root", database: "mysql")
client.query("SET NAMES utf8mb4")
client.query("DROP DATABASE IF EXISTS `repro`")
client.query("CREATE DATABASE `repro` DEFAULT CHARACTER SET `utf8mb4`")

client = Mysql2::Client.new(host: "127.0.0.1", username: "root", database: "repro")
client.query("SET NAMES utf8mb4")
client.query("CREATE TABLE t (id bigint NOT NULL AUTO_INCREMENT PRIMARY KEY, name varchar(10))")
client.query("INSERT INTO t (name) VALUES ('Test')")
22.times do
  client.query("INSERT INTO t (name) SELECT name FROM t")
end

p client.query("SELECT COUNT(*) AS count FROM t").first["count"]

sec = Benchmark.realtime do
  client.query("ALTER TABLE `t` CHANGE `name` `name` varchar(20) COLLATE utf8mb4_general_ci DEFAULT NULL")
end
p sec.round(3)
```

janko:
When I was pulling the `bitnami/mariadb:10.4` Docker image yesterday, it pulled version 10.4.27, and today it pulled 10.4.28, which appears to include a fix for this issue 🤘🏻

I verified that it's the same for 10.3.38, and my understanding is that 10.3 is the oldest supported version, so I don't think there is need to specify version after all 🙂

ankane:
fwiw, I'm not able to reproduce with 10.4.27, 10.4.6, or 10.3.7 either, which would also be supported.

janko:
Hmm, strange, I was able to reproduce it with the following script:

```rb
require "active_record"

ActiveRecord::Base.establish_connection("mysql2://root@127.0.0.1/testing")

ActiveRecord::Base.connection.create_table :items do |t|
  t.string :value, limit: 64
end

at_exit do
  ActiveRecord::Base.connection.drop_table :items
end

class Item < ActiveRecord::Base
end

Item.insert_all Array.new(1_000_000, { value: "abc123" })

ActiveRecord::Base.logger = Logger.new($stdout)

thread = Thread.new do
  ActiveRecord::Base.connection.change_column :items, :value, :string, limit: nil
end

sleep 0.5

Item.create(value: "blabla")

thread.join

# not necessary, just for comparison
ActiveRecord::Base.connection.change_column :items, :value, :string, limit: 64
```

For MariaDB 10.4.27, both increasing and decreasing column limit takes time proportional to the table size, and writes are blocked during that:

```sh
$ docker run --rm --name mariadb -e ALLOW_EMPTY_PASSWORD=yes -e MARIADB_DATABASE=testing -p 3306:3306 bitnami/mariadb:10.4.27
```
```
TRANSACTION (1.0ms)  BEGIN
(17917.9ms)  ALTER TABLE `items` CHANGE `value` `value` varchar(255) COLLATE utf8_general_ci DEFAULT NULL
Item Create (17402.1ms)  INSERT INTO `items` (`value`) VALUES ('blabla')
TRANSACTION (2.4ms)  COMMIT
(17623.7ms)  ALTER TABLE `items` CHANGE `value` `value` varchar(64) COLLATE utf8_general_ci DEFAULT NULL
(38.2ms)  DROP TABLE `items`
```

For MariaDB 10.4.28, increasing column limit is instantaneous regardless of the table size, and writes are probably not blocked (if they are, the window is very short anyway):

```sh
$ docker run --rm --name mariadb -e ALLOW_EMPTY_PASSWORD=yes -e MARIADB_DATABASE=testing -p 3306:3306 bitnami/mariadb:10.4.28
```
```
(5.6ms)  ALTER TABLE `items` CHANGE `value` `value` varchar(255) COLLATE utf8mb4_unicode_ci DEFAULT NULL
TRANSACTION (0.6ms)  BEGIN
Item Create (2.1ms)  INSERT INTO `items` (`value`) VALUES ('blabla')
TRANSACTION (1.0ms)  COMMIT
(12353.3ms)  ALTER TABLE `items` CHANGE `value` `value` varchar(64) COLLATE utf8mb4_unicode_ci DEFAULT NULL
(3.8ms)  DROP TABLE `items`
```

ankane:
Looks like it's all related to collation (the `bitnami/mariadb` image uses `utf8_general_ci` for 10.4.27 and `utf8mb4_unicode_ci` for 10.4.28). From what I can tell, `utf8_general_ci` reserves 3 bytes for each character in 10.4.x, while `utf8mb4_unicode_ci` reserves 4 bytes.

In the test script above, `varchar(64)` reserves 256 bytes with `utf8mb4_unicode_ci`, meaning it can be safely increased to the max, while `utf8_general_ci` reserves 192 bytes, so it can only be safely increased to 255 bytes.

Will push a fix to take collation into account.

janko:
That's really useful info, I think it will help us a lot in figuring out how to safely increase the column limit for a large table at work. Thanks for the help 🙏

ankane:
No problem, thanks for the report and repro script. Just pushed a new release that fixes the existing check and updated the docs.

A few other notes for future reference:

- The maxlen is set on the [charset](https://mariadb.com/kb/en/supported-character-sets-and-collations/) (collations belongs to a charset)
- Rails 5.2 generates `database.yml` with `encoding: utf8` while Rails 6+ uses `encoding: utf8mb4`
- Now that we know the cause, I didn't find a difference between MariaDB 10.4 and 10.5

janko:
> Now that we know the cause, I didn't find a difference between MariaDB 10.4 and 10.5

Yes, there is no difference as far as I know, I just happened to be testing on the latest 10.5.x version (currently 10.5.19), which probably had the changed collation you mentioned, and compared it with not-latest 10.4.x version at the time. When I tested the previous 10.5.x patch version (10.5.18), raising the limit triggered the table rewrite. Sorry for the initial confusion 😅

PR #218, opened 2023-03-03
Add add_column with index check

hlascelles:
When ActiveRecord adds a column with an index, it adds the column (safe) then adds the index concurrently. This locks the table, and is unsafe.

This PR adds a check for adding columns with an index.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## UNRELEASED
+
+- Added check for `add_column_with_index`
+
 ## 1.4.3 (2023-02-19)
 
 - Fixed check for `change_column` to account for charset with MySQL and MariaDB
--- a/README.md
+++ b/README.md
@@ -157,6 +157,38 @@ end
 
 See the next section for how to backfill.
 
+### Adding a column with an index
+
+#### Bad
+
+When adding a column with an index, two internal operations are performed: adding the column and adding the index. The first is safe, but the second is not.
+
+```ruby
+class AddSomeIndexedColumnToUsers < ActiveRecord::Migration[7.0]
+  def change
+    add_column :users, :some_column, :text, index: true
+  end
+end
+```
+
+#### Good
+
+Instead, add the column without an index, then add the index in a separate migration, using concurrent index creation.
+
+```ruby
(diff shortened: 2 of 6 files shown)


fatkodima:
Are you sure `add_column` accepts an `:index` option?

hlascelles:
Aha... Yes, if using [schema_plus_indexes](https://github.com/SchemaPlus/schema_plus_indexes#create-indexes-with-add_column-change_table):

"ActiveRecord supports the index: option to column definitions when creating table. SchemaPlus::Indexes extends that to work also with add_column and in change_table"

```ruby
add_column "tablename", "columnname", index: { ... }
```

OK, so what to do here. We were definitely lulled into an enhanced sense of security by `strong_migrations`, but it definitely wasn't its fault as this was a third party gem feature.

You are right, Rails does by default throw the `index: true` away silently.

I propose this still goes in with better a better error message that can tell you what is going on. Depending on the scenario:

"You have attempted to add an index when performing add_column, but this feature is not supported by Rails"
OR 
"You have attempted to add an index when performing add_column using `schema_plus_indexes`, but this feature is unsafe".

I understand there may be a concern for feature bleed (you can't be expected to cover every other gem), but maybe for the first reason alone (telling people they have mis-used an option that is confusingly allowed in other Rails migrations) it can be added?

fatkodima:
I am sure that @ankane will suggest you to use custom checks - https://github.com/ankane/strong_migrations#custom-checks.

fatkodima:
I would recommend to get rid of the mentioned gem and use the standard rails helpers. I just checked its readme and, honestly, do not see how it is will be useful or more convenient. Adding an index with the column still creates 2 sql queries, not 1 (probably someone expects to have only 1 and thus be faster?).

ankane:
Hey @hlascelles, thanks for the PR. As @fatkodima mentioned, I think it's best to use custom checks for this.

hlascelles:
Understood, we will do. Something like this as an initializer:

```ruby
    if method == :add_column && args[3].is_a?(Hash)
      index_args = args[3][:index]
      if index_args
        stop!(<<~MSG)
          Adding an index when adding a column is unsafe.
          Please add the index in a separate migration using `algorithm: :concurrently`.
        MSG
      end
    end
  end
```

Thanks!

PR #221, opened 2023-03-08
fix: pass through :name option in safe add_foreign_key

### LINKED ISSUE #220: add_foreign_key does not respect name when validating
If `safe_by_default` is enabled, `add_foreign_key` is broken into two steps: one of which adds the new key (un-validated), and one of which validates it.

Unfortunately, the `validate_foreign_key` call doesn't pass through the `name` parameter, so if you have multiple foreign keys pointing to the same table, the wrong one gets validated.

Roguelazer:
Fixes #220


ankane:
Thanks @Roguelazer!

Issue #180, opened 2022-04-04
Misunderstandings related to safe_by_default


khamusa:
First of all, thanks for this great work, It's a lifesaver :)

In our team at Oyster, we had some misunderstandings on an open PR setting safe_by_default to true. The documentation reads "Make operations safe by default" and lists 4 types of operations that are supposed to be affected. Which one of the following would be the correct interpretation?

- The gem will assume that you know what you're doing when performing those operations and not raise any errors (i.e., those checks will be skipped) -> turning `safe_by_default` on increases risk as the gem becomes permissive by default for these types of operations;
- The gem will change its behavior to a more strict enforcement policy where these types of operations will also raise exceptions (while the default  -> turning `safe_by_default` off decreases the risk, as the gem is permissive by default and this flag will make it restrictive.

I **believe** the correct interpretation is the first one, could you confirm?

Thank you! :raised_hands:

fatkodima:
I would say, it is neither 1 nor 2.

For this setting, the gem will change its behavior. It **will** perform the actual checks, but implicitly apply the mentioned 4 operations in a safe way.

For example, 
1. if you miss `algorithm: :concurrently`, it will not raise a warning, but add this config automatically for you
2. when you add a foreign key, it will not raise a warning, but add it in 2 steps (adding invalid and validate) for you
3. etc

ankane:
Hey @khamusa, in terms of risk, it shouldn't really increase or decrease it (Strong Migrations will raise an error for dangerous operations in either case). For the add index case, the option allows you to write:

```ruby
add_index :users, :some_column
```

which would normally raise an error, and instead, it'll automatically convert it to the safe version (as @fatkodima mentioned)

```ruby
add_index :users, :some_column, algorithm: :concurrently
```

For each operation, it allows you to write code under the "Bad" section of the readme and converts it to the code under the "Good" section.

PR #229, opened 2023-06-20
Describe safe_by_default behaviour explicitly

tomchipchase:
I wasn't 100% sure what the behaviour of this option was from the current description, and it looks like it [has been asked before](https://github.com/ankane/strong_migrations/issues/180), so I thought I'd add it to the readme for the next person.


ankane:
Thanks @tomchipchase! (and sorry for the delay)

PR #231, opened 2023-07-03
Adding temporary models as a preferred option for migration backfilling

panbanda:
In our continuous quest for sustainable and efficient codebase maintenance, it has become increasingly clear that our current practice of using User.unscoped for migrations in our legacy application, as suggested in our existing documentation, has some limitations. Over time, we have encountered multiple instances where this approach falls short, particularly for larger, more complex applications. Therefore, this pull request aims to propose a shift towards the adoption of temporary models for handling migrations.

Temporary models offer a few significant advantages that make them well-suited for our needs. Chiefly, they inherently offer forward compatibility, which is a crucial attribute for long-term application maintenance. As the application grows in size and complexity, this attribute becomes even more important, as it helps reduce the burden of navigating through the complex interrelationships between application models and migrations.

Moreover, temporary models allow for greater flexibility in managing migrations over time. Consider a scenario where a migration and model are created in an application. Fast forward three years, and suppose there's a need to remove that model. In our current setup, this would lead to a failure when trying to re-run migrations sequentially (in the absence of db:load_schema), which is not an ideal situation.

In contrast, by using temporary models, the model would remain available within the migration, allowing you to execute migrations sequentially, even if the related table gets removed at a later stage.

By moving to temporary models, we not only maintain our code's historical integrity, but also improve the flexibility and reliability of our application over time. This approach future-proofs our codebase and makes it easier to maintain and modify our application as it continues to evolve.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,14 +177,20 @@ Also, running a single query to update data can cause issues for large tables.
 
 #### Good
 
-There are three keys to backfilling safely: batching, throttling, and running it outside a transaction. Use the Rails console or a separate migration with `disable_ddl_transaction!`.
+There are three keys to backfilling safely: batching, throttling, and running it outside a transaction. Use the Rails console or a separate migration with `disable_ddl_transaction!`.  
+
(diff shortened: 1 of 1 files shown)


ankane:
Hi @panbanda, thanks for the PR, but I don't think temporary models are needed. See #78 for more info.

PR #234, opened 2023-07-25
Support rollback in "Setting NOT NULL on an existing column" example

TastyPi:
`validate_check_constraint` fails in the rollback because the constraint does not exist, and `remove_check_constraint` needs the expression in order to recreate the constraint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -605,11 +605,11 @@ For Rails 6.1, use:
 ```ruby
 class ValidateSomeColumnNotNull < ActiveRecord::Migration[7.0]
   def change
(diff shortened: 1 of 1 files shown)


ankane:
Thanks @TastyPi, merged in b0296ff38ae776d03f90758457ab9372478ca3de.

ankane:
Just realized approach in the error message is slightly better (since it won't validate the constraint), so updated in a5d6713a3dbd818cc5d9f1e78bfcdeb7af012da5.

PR #235, opened 2023-08-07
Clarify use of `unscoped` when backfilling

gabo-cs:
The use of `unscoped` for backfilling was introduced in https://github.com/ankane/strong_migrations/pull/77, which I think was really nice! However, there's no explanation for it in the main README. This PR introduces that so everyone knows why the `unscoped` is a good thing to keep in mind there.

Open to feedback, of course! LMK what you think! Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,7 @@ class BackfillSomeColumn < ActiveRecord::Migration[7.0]
   end
 end
 ```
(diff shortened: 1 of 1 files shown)


ankane:
Hi @gabo-cs, thanks for the PR, and sorry for the delay. However, it's not something I think is needed (users who are curious can look up the method to see what it does).

PR #238, opened 2023-08-27
Add foreign key options to safe_add_reference

chaadow:
This will handle for example `foreign_key: { to_table: :table_name, on_delete: :nullify }`


ankane:
Awesome, thanks @chaadow!

PR #239, opened 2023-08-28
feat: Add action and start_after to custom checks

barthez:
Hey!

Firstly, thank you verymuch for all the work with strong migrations, great work!

Recently we have added few custom checks.  We had to use `enable_check` and `check_enabled?` methods with custom check names so we are able to enable them only after given migration version. I believe that could be part of StrongMigrations, what do you think?

### DIFF
--- a/lib/strong_migrations.rb
+++ b/lib/strong_migrations.rb
@@ -63,8 +63,16 @@ module StrongMigrations
     @lock_timeout_limit
   end
 
-  def self.add_check(&block)
-    checks << block
+  def self.add_check(only_action = nil, start_after: nil, &block)
+    check_id = ["__custom_check", *only_action, block.hash.to_s(16)].join("_").to_sym
+    enable_check(check_id, start_after: start_after)
+
+    checks << proc do |action, args|
+      next unless StrongMigrations.check_enabled?(check_id, version: version)
+      next unless only_action.nil? || action == only_action
+
+      instance_exec(action, args, &block)
+    end
   end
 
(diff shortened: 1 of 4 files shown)


lib/strong_migrations.rb
```diff
@@ -63,8 +63,16 @@ def self.lock_timeout_limit
     @lock_timeout_limit
   end
 
-  def self.add_check(&block)
-    checks << block
+  def self.add_check(only_action = nil, start_after: nil, &block)
```
fatkodima, line 66: The `&block` already accepts operation name as an argument, so I think the implementation can be revised and simplified. 
For example, to something like this
https://github.com/fatkodima/online_migrations/blob/5384767ccd3f8246b123780525d9a9763251ac38/lib/online_migrations/config.rb#L207-L260

barthez: Yes, true that. Although, I believe the argument for `add_check` conveys the information a bit better than first line in a check body that says which action this check is about. Even though the code do exactly the same, but implicitly. 

```ruby
StrongMigrations.add_check(:remove_index) do 
  ...
end
```

vs

```ruby
StrongMigrations.add_check do |action, args|
  next unless action == :remove_index

  ...
end

But, yeah, if you really don't like it, I can remove, my main focus was on `start_after`.

**Comment by ankane** [comment]
Hi @barthez, thanks for the PR. It's already possible to check the action and migration version in a custom check, so I don't think additional options are needed.

```ruby
  next if version < 1000
  # ...
end
```

PR #241, opened 2023-09-07
Drop invalid index in safe_add_index

dgsuarez:
Hi! Saw this TODO in the code, we've been running something similar in our codebase, so better contribute it back :) 

Allow `safe_add_index` to be seamlessly retried.

When the `add_index...concurrently` fails, the index is left in Postgres with `INVALID` status.

The new code drops the index if it's found and invalid. With this, if the migration fails because of a lock timeout for example, on retry it can work. Without it the retries always fail until you manually drop the index.

### DIFF
--- a/lib/strong_migrations/adapters/postgresql_adapter.rb
+++ b/lib/strong_migrations/adapters/postgresql_adapter.rb
@@ -169,6 +169,22 @@ module StrongMigrations
         rows.empty? || rows.any? { |r| r["provolatile"] == "v" }
       end
 
+      def invalid_index?(index_name)
+        query = <<~SQL
+          SELECT
+            pg_class.relname
+          FROM
+            pg_class
+          INNER JOIN
+            pg_index ON pg_index.indexrelid = pg_class.oid
+          WHERE
+            pg_index.indisvalid = false AND
+            pg_class.relname = #{connection.quote(index_name)}
+        SQL
+
+        select_all(query.squish).any?
+      end
+
       private
 
       def set_timeout(setting, timeout)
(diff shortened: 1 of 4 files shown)


dgsuarez:
Small update to fix the errors in pre-6.1 rails versions

chaadow:
@ankane For my case, this pull request is helpful because when indices are `INVALID` rails does not add the `CREATE UNIQUE INDEX` inside `db/structure.sql`

which in CI would not generate a proper database and make the tests fail ( and in a team of many developers, this creates issues where `structure.sql` differ a lot from one developer ( who does not have an invaild index ) to another who has

the solution is to manually remove the index from psql or in some other way, but it would be great to have this PR merged to automatically handle the case

Thanks!

ankane:
Hi @dgsuarez, thanks for the PR, and sorry for the delay. Merged a version of this in the commit above (requires Active Record 7.1+, but may decide to support earlier versions before the release).

ankane:
Spent some more time on this and think it's better to call `REINDEX INDEX CONCURRENTLY` (when possible), since it only requires acquiring a lock once.

ankane:
It looks like `REINDEX INDEX CONCURRENTLY` leaves behind another invalid index if it fails, so reverted back to `DROP INDEX CONCURRENTLY`.

Also, this seems like a good option to have independent of `safe_by_default`, so added experimental support for:

```ruby
StrongMigrations.remove_invalid_indexes = true
```

MinaGeorgi:
Hello @ankane and thanks for adding the extra check. If it can be done, it would be great to have it also for ActiveRecord <7.1

PR #243, opened 2023-09-26
Add changelog uri in metadata

MaximeD:
Rubygems allows to specify [metadata](https://guides.rubygems.org/specification-reference/#metadata).

This patch adds the `changelog_uri` to help others find where the CHANGELOG is.


ankane:
Hi @MaximeD, thanks for the PR, but it's not something I'd like to add.

PR #259, opened 2024-03-26
Fix a typo

ydah:
s/Contraint/Constraint/


ankane:
Nice catch, thanks @ydah!

PR #260, opened 2024-03-26
Added dependabot for GitHub Actions

ydah:
How about using dependabot in this way?
When this works, a pull request is created as follows:
- ruby/bigdecimal#242

It was created with reference to the following:
- ruby/csv#265


ankane:
Thanks for another PR, but it's not something I'd like to add.

Issue #129, opened 2020-09-18
Conflict when using with toptal/chewy Gem.


leduytung:
Hi,

I having trouble when using the strong_migration gem with the **toptal/chewy gem**.
Both 2 gems rewrite the method migrate of Active Support, lead to an infinite loop in the migrate function when I run **rake db:migrate**
Here are the error messages:
`
....
/home/meowth/.rvm/gems/ruby-2.5.3/gems/chewy-5.1.0/lib/chewy/railtie.rb:32:in `block in migrate_with_chewy'
/home/meowth/.rvm/gems/ruby-2.5.3/gems/chewy-5.1.0/lib/chewy/strategy.rb:70:in `wrap'
/home/meowth/.rvm/gems/ruby-2.5.3/gems/chewy-5.1.0/lib/chewy.rb:201:in `strategy'
/home/meowth/.rvm/gems/ruby-2.5.3/gems/chewy-5.1.0/lib/chewy/railtie.rb:32:in `migrate_with_chewy'
/home/meowth/.rvm/gems/ruby-2.5.3/gems/strong_migrations-0.3.1/lib/strong_migrations/migration.rb:13:in `migrate'
/home/meowth/.rvm/gems/ruby-2.5.3/gems/chewy-5.1.0/lib/chewy/railtie.rb:32:in `block in migrate_with_chewy'
/home/meowth/.rvm/gems/ruby-2.5.3/gems/chewy-5.1.0/lib/chewy/strategy.rb:70:in `wrap'
/home/meowth/.rvm/gems/ruby-2.5.3/gems/chewy-5.1.0/lib/chewy.rb:201:in `strategy'
/home/meowth/.rvm/gems/ruby-2.5.3/gems/chewy-5.1.0/lib/chewy/railtie.rb:32:in `migrate_with_chewy'
/home/meowth/.rvm/gems/ruby-2.5.3/gems/strong_migrations-0.3.1/lib/strong_migrations/migration.rb:13:in `migrate'
/home/meowth/.rvm/gems/ruby-2.5.3/gems/chewy-5.1.0/lib/chewy/railtie.rb:32:in `block in migrate_with_chewy'
/home/meowth/.rvm/gems/ruby-2.5.3/gems/chewy-5.1.0/lib/chewy/strategy.rb:70:in `wrap'
/home/meowth/.rvm/gems/ruby-2.5.3/gems/chewy-5.1.0/lib/chewy.rb:201:in `strategy'
[...]

PR #270, opened 2024-08-28
Fix SystemStackError in schema_dumper.rb in combination with activerecord-multi-tenant

Laykou:
When using https://github.com/citusdata/activerecord-multi-tenant gem, it also overrides the SchemaDumper. This gem uses alias method instead of `super`.

Reference:
https://github.com/citusdata/activerecord-multi-tenant/blob/0ac43aa98334a29ed3e5b33e0bcc5ee281f97254/lib/activerecord-multi-tenant/migrations.rb#L86

strong_migrations in combination with this gem created following issue:

```
SystemStackError: stack level too deep (SystemStackError)
.../gems/strong_migrations-1.8.0/lib/strong_migrations/schema_dumper.rb:6:in `new'
.../gems/strong_migrations-1.8.0/lib/strong_migrations/schema_dumper.rb:6:in `initialize'
.../gems/activerecord-multi-tenant-2.4.0/lib/activerecord-multi-tenant/migrations.rb:93:in `initialize'
.../gems/strong_migrations-1.8.0/lib/strong_migrations/schema_dumper.rb:6:in `initialize'
.../gems/activerecord-multi-tenant-2.4.0/lib/activerecord-multi-tenant/migrations.rb:93:in `initialize'
.../gems/strong_migrations-1.8.0/lib/strong_migrations/schema_dumper.rb:6:in `initialize'
```

I figured out using the `prepended` with the `alias` method instead avoided the issue in overriding the `initialize` method by multiple gems

### DIFF
--- a/lib/strong_migrations/schema_dumper.rb
+++ b/lib/strong_migrations/schema_dumper.rb
@@ -1,9 +1,16 @@
 module StrongMigrations
   module SchemaDumper
-    def initialize(connection, *args, **options)
-      return super unless StrongMigrations.alphabetize_schema
+    extend ActiveSupport::Concern
 
-      super(WrappedConnection.new(connection), *args, **options)
+    prepended do
+      alias original_initialize initialize
+
+      def initialize(connection, *args, **options)
+
+        return original_initialize(connection, *args, **options) unless StrongMigrations.alphabetize_schema
+
+        original_initialize(WrappedConnection.new(connection), *args, **options)
+      end
     end
   end
 


ankane:
Hi @Laykou, thanks for the PR, but I think it's better for conflicting gems to switch to `Module#prepend` (#129).

Laykou:
@ankane Thank you, I may not be that fluent in this "Ruby overriding" world. So you're saying using:

```ruby
ActiveRecord::SchemaDumper.prepend(StrongMigrations::SchemaDumper)
```

like you do in this gem is preferred/better option rather than monkey pathing:

```ruby
module ActiveRecord
  class SchemaDumper
    alias initialize_without_citus initialize

    def initialize(connection, options = {})
      # ... calling 
    end
  end
end
```

that is done in the other gem? And the root cause of the issue is this monkey patching.
Calling `super` when using _prepend_ is the correct way, right?

Thank you for your insights

ankane:
Yes, I think `Module#prepend` is the better pattern.

PR #272, opened 2024-09-01
Fix add_check_constraint  when raising an error in safe mode

chaadow:
Followup to https://github.com/ankane/strong_migrations/pull/271

When adding a CHECK constraint and safe mode is enabled:

- we add a NOT VALID check constraint
- then commit the transaction
- and finally validate the check constraint in another migration.

If the validation step fails, this leaves a NOT VALID check constraint in the database.
If we retry after having cleaned up the database, the creation of the NOT VALID check constraint will fail because it will already be in the DB.

This commit makes the migration safely retriable by checking if said check constraint already exists, and if so, skips its creation again.

### DIFF
--- a/lib/strong_migrations/adapters/postgresql_adapter.rb
+++ b/lib/strong_migrations/adapters/postgresql_adapter.rb
@@ -127,7 +127,7 @@ module StrongMigrations
         safe
       end
 
-      def constraints(table_name)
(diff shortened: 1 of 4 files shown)


ankane:
Hey @chaadow, thanks for the PR. Merged a version in the commit above. I'm not sure about trying to clean up if the migration fails (as it'll introduce more locking), so excluding it for now.

PR #273, opened 2024-09-01
Make add_foreign_key retriable when safe_by_default is enabled

chaadow:
When adding a foreign key and safe mode is enabled:
- We create a NOT VALID foreign key
- Then we commit the transaction
- Finally we validate the foreign key constraint

if the validation fails, this leaves a NOT VALID foreign key in the database.
If we retry after having cleaned up the database, the creation of the NOT VALID foreign key constraint will fail because it will already be in the DB.

This commit makes the migration safely retriable by checking if said foreign key constraint already exists, and if so, skips its creation again.

When validation of foreign key fails, we also remove the NOT VALID foreign key constraint as to not leave the database in an unstable state.


lib/strong_migrations/safe_methods.rb
```diff
@@ -45,15 +45,25 @@ def safe_add_reference(table, reference, *args, **options)
     end
 
     def safe_add_foreign_key(from_table, to_table, *args, **options)
+      validate_options = remove_options = options.slice(:column, :name)
       @migration.reversible do |dir|
         dir.up do
-          @migration.add_foreign_key(from_table, to_table, *args, **options.merge(validate: false))
+          # https://github.com/rails/rails/blob/main/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb#L1154C96-L1154C120
+          # if_not_exists does not check again `name` option.
```
chaadow, line 52: Until a fix is done/merged on Rails, I believe we still need to leave this.

This was detected by `AddForeignKeyName` test class, as it creates 2 foreign keys on the same column but with different names ( It's quite rare I think for users of this gem to encounter this case but nonetheless, it's more robust and makes tests pass )

ankane:
Thanks @chaadow, merged a version in the commit above.

PR #274, opened 2024-09-13
Fixed check_add_column

saviokmua:
I tried to add a test boolean field

```ruby
class AddNewFieldToUsers < ActiveRecord::Migration[7.1]
  def change
    add_column :users, :new_field, :boolean, default: false, null: false
  end
end
```

and checker of strong migrations didn't work for my case

PostgreSQL 14.13
Rails 7.1.3.4
ruby 3.3.0

I tried to debug why this happened and came across the fact that `default` ==` false` (default is not Proc)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@
 /pkg/
 /spec/reports/
 /tmp/
+/.idea
+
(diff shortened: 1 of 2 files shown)


ankane:
Hi @saviokmua, that operation isn't dangerous (which is why it didn't raise an error).

saviokmua:
Hi @ankane. hm.... very strange. I have a problem with locking the user table when performing a migration

PR #278, opened 2024-10-15
Add CHANGELOG to gem metadata

leoarnold:
Dependency updaters like Dependabot, Depfu, or RenovateBot can link the changelog of a gem in their PRs if the changelog URI is specified in the gem's spec.


ankane:
Hi @leoarnold, thanks for the PR, but see #243.

PR #280, opened 2024-10-20
Add metadata to gemspec

maxjacobson:
Hello!

Just noticed while browsing rubygems that strong_migrations doesn't have a changelog or source code link in the "links" section for the gem:

CleanShot 2024-10-20 at 14 21 31@2x

From: https://rubygems.org/gems/strong_migrations

If the gemspec provides this [metadata](https://guides.rubygems.org/specification-reference/#metadata), rubygems will add those links.

Personally I find those useful, so I thought I'd open this PR to suggest adding them. WDYT?

### DIFF
--- a/strong_migrations.gemspec
+++ b/strong_migrations.gemspec
@@ -16,4 +16,9 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = ">= 3.1"
 
   spec.add_dependency "activerecord", ">= 6.1"
+
+  spec.metadata = {
+    "changelog_uri"     => "https://github.com/ankane/strong_migrations/blob/v#{StrongMigrations::VERSION}/CHANGELOG.md",
+    "source_code_uri"   => "https://github.com/ankane/strong_migrations",
+  }
 end


strong_migrations.gemspec
```diff
@@ -16,4 +16,9 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = ">= 3.1"
 
   spec.add_dependency "activerecord", ">= 6.1"
+
+  spec.metadata = {
+    "changelog_uri"     => "https://github.com/ankane/strong_migrations/blob/v#{StrongMigrations::VERSION}/CHANGELOG.md",
+    "source_code_uri"   => "https://github.com/ankane/strong_migrations",
```
maxjacobson, line 22: I was tempted also to add a `documentation_uri` which links to `"https://github.com/ankane/strong_migrations/blob/v#{StrongMigrations::VERSION}/README.md"`. The default documentation link goes to https://www.rubydoc.info/gems/strong_migrations/2.0.1, which also works fine. But as a user of the gem, I'm less interested in the internals. I just want the content in the README.

fatkodima:
https://github.com/ankane/strong_migrations/pull/278

> Hi MaximeD, thanks for the PR, but it's not something I'd like to add.

Guess why.

maxjacobson:
Ah, whoops, I didn't check if this had been proposed before. I suppose I'll close.

> Guess why.

Not sure! Why?

fatkodima:
> Not sure! Why?

Nobody knows 😄 But it was proposed many times and there is no reason to include it.

PR #294, opened 2025-03-03
Disable `safe_by_default` on existing migrations

sinsoku:
When `safe_by_default` is enabled, an error occurs in `db:migrate` if
existing migrations has `change_column_null` with a default value.

refs: https://github.com/ankane/strong_migrations/blob/8e9baaa05c35fc9fdf206160c00b309c6e3a8bb6/lib/strong_migrations/safe_methods.rb#L83

To avoid this error, disable `safe_by_default` on existing migrations.
Not required for migrations created before the strong_migrations gem is installed.

### DIFF
--- a/lib/strong_migrations/checker.rb
+++ b/lib/strong_migrations/checker.rb
@@ -41,7 +41,7 @@ module StrongMigrations
       set_timeouts
       check_lock_timeout
 
-      if !safe? || safe_by_default_method?(method)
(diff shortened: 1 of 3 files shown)


ankane:
Hi @sinsoku, thanks for reporting and the PR! Merged a slightly modified version in the commit above that only looks at the `start_after` option.

PR #302, opened 2025-06-20
support transaction_timeout (introduced in PG17)

quentindemetz:
PG17 supports [transaction_timeout](https://postgresqlco.nf/doc/en/param/transaction_timeout/). Our application has set a relatively short value for this, as we have for statement_timeout. 

StrongMigrations changes the statement timeout to a large value. This is useless if the transaction_timeout itself remains low:

> The limit applies both to explicit transactions (started with BEGIN) and to an implicitly started transaction corresponding to a single statement.

—> StrongMigrations needs to override it to a larger value for migrations to pass.

I have not added any tests so far. Please let me know if there is something specific you would like me to test.


ankane:
Great, thanks!

quentindemetz:
Thank you for the fast turnaround time 🙏🏻 . Refreshing!

PR #303, opened 2025-07-18
Analyze when add_reference creates an index

rylwin:
Just like with add_index, when add_reference adds an index it's
important to ANALYZE the table to ensure stats are up-to-date.

### DIFF
--- a/lib/strong_migrations/checker.rb
+++ b/lib/strong_migrations/checker.rb
@@ -111,11 +111,16 @@ module StrongMigrations
         end
 
       # outdated statistics + a new index can hurt performance of existing queries
-      if StrongMigrations.auto_analyze && direction == :up && method == :add_index
+      if StrongMigrations.auto_analyze && direction == :up && migration_adds_an_index?(method, *args)
         adapter.analyze_table(args[0])
       end
 
       result
+    end
+
+    def migration_adds_an_index?(method, *args)
+      options = args.last
+      method == :add_index || (method == :add_reference && options[:index])
     end
 
     def perform_method(method, *args)
--- a/test/add_reference_test.rb
+++ b/test/add_reference_test.rb
@@ -57,4 +57,20 @@ class AddReferenceTest < Minitest::Test
       assert_argument_error AddReferenceExtraArguments
     end
   end
+
+  def test_auto_analyze_with_index
+    StrongMigrations.auto_analyze = true
+    logs = capture_logs { assert_safe AddReferenceConcurrently }
+    assert_includes logs, "ANALYZE"
+  ensure
+    StrongMigrations.auto_analyze = false
+  end
+
+  def test_auto_analyze_skips_analyze_without_index
+    StrongMigrations.auto_analyze = true
+    logs = capture_logs { assert_safe AddReferenceNoIndex }
+    refute logs.include?("ANALYZE")
+  ensure
+    StrongMigrations.auto_analyze = false
+  end
 end


rylwin:
Thanks for taking a look. FWIW, we had this exact situation cause downtime for us last week, which was solved by manually ANALZYE-ing the newly added (all null) column

rylwin:
Hi @ankane, happy to provide more details.

Last week we added a new foreign key column to a table with 300M+ rows. The migrations were (in three separate migration files):

```
add_reference(
  :activities,
  :multiple_community_parent_activity,
  index: { algorithm: :concurrently },
)
add_foreign_key(
  :activities,
  :activities,
  column: :multiple_community_parent_activity_id,
  validate: false
)
validate_foreign_key(
  :activities,
  column: :multiple_community_parent_activity_id,
)
```

The column was filled with NULL values. We started seeing simple lookup queries in the form of:

```
select * from activities where multiple_community_parent_activity_id IN (?, ?, ?...)
```

result in Seq Scans over the 300M row table. This quickly led to our database becoming overloaded with work.

Only once we ran `ANALYZE activities (multiple_community_parent_activity_id)` did PostgreSQL begin using the index.

We're running PostgreSQL 14.13 via AWS RDS.

I put together a set of scripts that reproduces the situation: https://github.com/WelcomeHome-Software/postgres-add-reference-with-index-needs-analyze. On my machine, these scripts reliably result in Seq Scan up until ANALYZE is run, at which point PostgreSQL starts using the index.

Happy to provide any other details that might be helpful.

ankane:
Awesome, will take a look.

ankane:
Looks like it can be an issue at 100M rows.

100M rows before analyze

```text
                                                             QUERY PLAN                                                              
-------------------------------------------------------------------------------------------------------------------------------------
 Gather  (cost=1000.00..1347982.33 rows=1000000 width=30) (actual time=4582.122..4590.467 rows=0 loops=1)
   Workers Planned: 2
   Workers Launched: 2
   ->  Parallel Seq Scan on activities  (cost=0.00..1246982.33 rows=416667 width=30) (actual time=4557.171..4557.172 rows=0 loops=3)
         Filter: (parent_id = ANY ('{1,2}'::integer[]))
         Rows Removed by Filter: 33333333
 Planning Time: 3.045 ms
 JIT:
   Functions: 6
   Options: Inlining true, Optimization true, Expressions true, Deforming true
   Timing: Generation 0.748 ms, Inlining 311.155 ms, Optimization 50.501 ms, Emission 42.289 ms, Total 404.693 ms
 Execution Time: 4618.549 ms
```

100M rows after analyze

```text
                                                              QUERY PLAN                                                               
---------------------------------------------------------------------------------------------------------------------------------------
 Index Scan using idx_activities_parent_id on activities  (cost=0.57..13.16 rows=1 width=30) (actual time=0.856..0.856 rows=0 loops=1)
   Index Cond: (parent_id = ANY ('{1,2}'::integer[]))
 Planning Time: 0.916 ms
 Execution Time: 0.919 ms
```

10M rows before analyze (not an ideal plan, but likely fine)

```text
                                                               QUERY PLAN                                                                
-----------------------------------------------------------------------------------------------------------------------------------------
 Bitmap Heap Scan on activities  (cost=1119.87..78812.61 rows=100000 width=30) (actual time=0.043..0.044 rows=0 loops=1)
   Recheck Cond: (parent_id = ANY ('{1,2}'::integer[]))
   ->  Bitmap Index Scan on idx_activities_parent_id  (cost=0.00..1094.87 rows=100000 width=0) (actual time=0.038..0.038 rows=0 loops=1)
         Index Cond: (parent_id = ANY ('{1,2}'::integer[]))
 Planning Time: 0.834 ms
 Execution Time: 0.148 ms
```

Thanks for putting together the great repro! Merged a version of this PR in the commit above.

rylwin:
Awesome -- thanks for taking the time to dig into this one. 

We've been using strong_migrations since 2022 and it's helped us avert countless migration issues. Thanks for maintaining this gem

PR #306, opened 2025-09-09
Check pending migrations for strong_migrations violations

kaka-ruto:
This PR

  - Adds a new `strong_migrations:check` rake task that validates all pending migrations against StrongMigrations rules without applying them
  - Reports which specific migrations fail validation with detailed error messages and counts
  - Exits with appropriate codes (0 for success, 1 for failures) for CI/CD integration

  Usage

  # Check all pending migrations
  bundle exec rake strong_migrations:check

  # Example output - no pending migrations
  No pending migrations found to check

  # Example output - all migrations pass
  All 3 pending migrations are safe!

  # Example output - some migrations fail
  Strong Migrations found issues with 2 of 3 pending migrations:

  20240101000001_AddIndex
     Adding an index non-concurrently blocks writes. Instead, use:
     class AddIndex < ActiveRecord::Migration[7.0]
       disable_ddl_transaction!
       def change
         add_index :users, :email, algorithm: :concurrently
       end
     end

  20240101000002_RemoveColumn
     Removing a column can break code...

  Run `bundle exec rails g strong_migrations:install` to configure.


ankane:
Hi @kaka-ruto, thanks for the PR. However, I don't see a strong reason to add this functionality.

PR #309, opened 2025-10-30
Enhance error messages column removal

alexeyltd:
## Context

We had an incident in our app because an engineer didn’t follow the guide correctly. I’d like to make it more straightforward.

## Implementation 
Enhanced error messages for column removal migrations with detailed instructions and warnings to prevent downtime.
Clarified the two-step process for safely removing columns in ActiveRecord.


ankane:
Hi @alexeyltd, the current message seems fine, but you can [customize it](https://github.com/ankane/strong_migrations?tab=readme-ov-file#custom-messages) for your application.

PR #317, opened 2026-03-26
Add note about safety_assured not carrying over during rollback

### LINKED ISSUE #284: Followed advice; rollback still won't run
Trying to rollback this:

```
  def change
    safety_assured do
      add_column :tournaments, :livestream_required, :boolean, default: false, null: false
      add_column :tournament_templates, :livestream_required, :boolean, default: false, null: false
    end
  end
```

I am advised:

```
Active Record caches attributes, which causes problems
when removing columns. Be sure to ignore the column:

class TournamentTemplate < ApplicationRecord
  self.ignored_columns = ["livestream_required"]
end

Deploy the code, then wrap this step in a safety_assured { ... } block.

class AddLivestreamRequired < ActiveRecord::Migration[7.0]
  def change
    safety_assured { remove_column :tournament_templates, :livestream_required, :boolean, default: false, null: false }
  end
end
/Users/e86037/.asdf/installs/ruby/3.2.2/bin/bundle:25:in `load'
/Users/e86037/.asdf/installs/ruby/3.2.2/bin/bundle:25:in `<main>'

Caused by:

Active Record caches attributes, which causes problems
when removing columns. Be sure to ignore the column:

class TournamentTemplate < ApplicationRecord
  self.ignored_columns = ["livestream_required"]
end

Deploy the code, then wrap this step in a safety_assured { ... } block.

class AddLivestreamRequired < ActiveRecord::Migration[7.0]
  def change
    safety_assured { remove_column :tournament_templates, :livestream_required, :boolean, default: false, null: false }
  end
end
```

Note that I got the advice twice for some reason. Perhaps because there are two tables involved, but both advice only mentions one of them.

I have these lines in my models:
```
class TournamentTemplate < ApplicationRecord
  self.ignored_columns = ["livestream_required"]
…
class Tournament < ApplicationRecord
  self.ignored_columns = ["livestream_required"]
```

yet the migration still refuses to run.

55728:
Fixes #284.

When using `safety_assured` with `add_column` inside a `change` method, the `safety_assured` context does not carry over when Rails reverts the migration. This can be confusing when `check_down` is enabled, as the rollback hits the `remove_column` check with no indication of why `safety_assured` didn't help.

Building on d15e3fa which improved the error output formatting for rollbacks, this adds a brief note to the message content itself suggesting the use of `up` and `down` methods instead.

### DIFF
--- a/lib/strong_migrations/error_messages.rb
+++ b/lib/strong_migrations/error_messages.rb
@@ -89,7 +89,10 @@ class %{migration_name} < ActiveRecord::Migration%{migration_suffix}
   def change
     safety_assured { %{command} }
   end
-end",
+end
+
+Note: safety_assured does not carry over when rolling back a change method.
+Use up and down methods instead if you need to roll back safely.",
 
     rename_column:
 "Renaming a column that's in use will cause errors


ankane:
Hi @55728, thanks for the PR, but I think the message will be noise for most users.

55728:
Thanks for taking a look!

PR #318, opened 2026-04-09
Add metadata to gemspec

jgarber623:
Configures links that will appear on this gem's page on RubyGems.org when new versions of the gem are released. See the [specification reference](https://guides.rubygems.org/specification-reference/#metadata) for more.

If nothing else, I'd like to see the addition of `changelog_uri` merged as its presence is quite helpful when working through gem updates and navigating via RubyGems.org.

Note that the duplicative `homepage_uri` accounts for [this issue](https://github.com/rubygems/rubygems.org/issues/4924).

Thanks for considering this change!

### DIFF
--- a/strong_migrations.gemspec
+++ b/strong_migrations.gemspec
@@ -15,5 +15,11 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = ">= 3.3"
 
+  spec.metadata["bug_tracker_uri"] = "#{spec.homepage}/issues"
+  spec.metadata["changelog_uri"] = "#{spec.homepage}/blob/v#{spec.version}/CHANGELOG.md"
+  spec.metadata["documentation_uri"] = "https://rubydoc.info/gems/#{spec.name}/#{spec.version}"
+  spec.metadata["homepage_uri"] = spec.homepage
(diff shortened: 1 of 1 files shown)


strong_migrations.gemspec
jgarber623, line 1: Alternatively, this could be expressed as a single Hash:

```ruby
spec.metadata = {
  "bug_tracker_uri" => "#{spec.homepage}/issues",
  # etc. etc. etc.
}
```

Note that Bundler/RubyGems expects the `"key" => "value"` style.

fatkodima:
You can find 10 similar PRs, which were closed previously with no explanation **why** it is not needed 😅 I've been dying to know the reason for multiple years now.

jgarber623:
Oh, jeez, okay! Thanks for that bit of historical detail.

My apologies to the maintainer if these proposed changes are undesirable. Certainly not my intention.

jgarber623:
Closing. See: https://github.com/ankane/strong_migrations/pull/243#issuecomment-1880247379

fatkodima:
> thanks for the PR, but it's not something I'd like to add.

People open and close their PRs again and again, but still no one knows the reason. I don't understand why this is a bad addition. When browsing rubygems.org, I do not want to do extra work to find the project's changelog instead of just clicking the link.

PR #320, opened 2026-05-05
Added safe_change_table to inspect operations inside change_table blocks

nsarno:
## Summary

Adds `safe_change_table`, a checked alternative to `safety_assured { change_table … }`. Each operation inside the block is validated through the same per-operation checks Strong Migrations applies at the top level. If any operation would be unsafe, the migration is rejected with the same error message as the equivalent top-level call.

```ruby
class AddNamesToUsers < ActiveRecord::Migration[8.1]
  def change
    safe_change_table :users, bulk: true do |t|
      t.string :first_name
      t.string :last_name
      t.index [:first_name, :last_name]
    end
  end
end
```

When `StrongMigrations.safe_by_default = true`, plain `change_table` is automatically rewritten to `safe_change_table`.

## Motivation

Today, `change_table` is rejected outright and the only workaround is `safety_assured { change_table … }`, which has two undesirable properties:

1. It disables every check (including custom checks registered via `StrongMigrations.add_check`), even for operations that would have been individually safe.
2. It penalizes the most efficient pattern for multi-column changes on MySQL/MariaDB. Users want `change_table :t, bulk: true` to combine multiple ALTERs into a single statement, but the only way to use it today is via `safety_assured`.

`safe_change_table` gives a checked equivalent. The bulk ALTER still runs as one statement, but every operation inside the block is validated first.

## How it works

1. The block is yielded to a `Table` proxy backed by `ActiveRecord::Migration::CommandRecorder`, capturing each operation as `(method, args, block)` without issuing any SQL.
2. Each recorded command is dispatched through the existing `Checker#perform` method, so it goes through the same per-operation checks (`check_add_column`, `check_add_index`, `check_change_column`, `check_remove_column`, `check_add_reference`, etc.) and any custom checks registered via `StrongMigrations.add_check`.
3. If all checks pass, `connection.change_table(table_name, **options, &block)` is called inside `safety_assured` so the real changes are applied (preserving `bulk: true` and any other options).

The block runs twice: once for inspection, once for execution. This is the same pattern Active Record already uses for `change_table … bulk: true`, so blocks that work with `bulk: true` today work with `safe_change_table` unchanged. Non-DDL side effects in the block run twice; this is documented in the README.

## Backwards compatibility

- Plain `change_table` (without the `safe_` prefix) is unchanged. It still raises, preserving the existing guardrail.
- `safety_assured { change_table … }` still works.
- Users who don't enable `safe_by_default` see no behavior change.

## Tests

New `test/change_table_test.rb` covers:

- Plain `change_table` still raises (regression).
- `safe_change_table` with safe operations applies the changes and is reversible.
- `safe_change_table` containing an unsafe operation raises and applies nothing.
- Custom checks (`StrongMigrations.add_check`) fire for operations inside the block.
- `safe_change_table :t, bulk: true` issues a single bundled `ALTER TABLE` on MySQL/MariaDB.
- `safe_change_table` without a block raises `ArgumentError`.

`test/safe_by_default_test.rb` is extended to cover the `change_table` rewrite (safe ops apply, unsafe ops still raise).

The full suite passes on PostgreSQL, MySQL, and Trilogy.

## Docs

- New section in `README.md`.
- `change_table` listed under "Safe by Default".
- `CHANGELOG.md` entry under a new `Unreleased` section.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Unreleased
+
+- Added `safe_change_table` to inspect operations inside `change_table` blocks
+- Made plain `change_table` safe by default when `StrongMigrations.safe_by_default = true`
+
 ## 2.7.0 (2026-04-25)
 
 - Added check for `add_foreign_key` with MySQL and MariaDB
--- a/README.md
+++ b/README.md
@@ -423,6 +423,31 @@ class ExecuteSQL < ActiveRecord::Migration[8.1]
 end
 ```
 
+### Inspecting change_table blocks
+
+Strong Migrations rejects plain `change_table` because it cannot see what happens inside the block. Use `safe_change_table` instead. Each operation in the block is validated as if it were a top-level call:
+
+```ruby
+class AddNamesToUsers < ActiveRecord::Migration[8.1]
+  def change
+    safe_change_table :users, bulk: true do |t|
+      t.string :first_name
+      t.string :last_name
+      t.index [:first_name, :last_name]
+    end
+  end
+end
+```
+
+If any operation in the block would be unsafe, the migration is rejected with the same error message as the equivalent top-level call.
+
+Note: the block is executed twice
+
+- once to capture operations for inspection (no SQL is issued)
+- once by the real `change_table` to apply the changes.
+
+So any non-DDL side effects in the block run twice. This matches Active Record’s existing behavior for `change_table` with `bulk: true`.
+
 ### Backfilling data
 
 Note: Strong Migrations does not detect dangerous backfills.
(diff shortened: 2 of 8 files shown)


ankane:
Hi @nsarno, thanks for the PR, but I don't think it's worth the additional complexity right now (for safe operations, I don't think combining changes is that much more efficient, but please share if you have numbers for specific migrations).

nsarno:
@ankane Thanks for your answer! And no worries, I understand the concern.

Just for context, I was actually hoping to benefit from the atomicity of `change_table bulk: true` when using MySQL, without having to rely on `safety_assured`.

Currently we're forced to choose between writing a non-atomic migration or using `safety_assured`, and we're trying to stay clear of both as a general rule within our company. That's a specific problem to MySQL because migrations don't run within a transaction like they are in Postgres.

ankane:
If you use separate migrations for each change, individual migrations can still be atomic.

nsarno:
Yes, that's the solution I'm using now.

PR #322, opened 2026-09-22
Fix alphabetize_schema on Rails 8.2

zarqman:
In rails/rails#58494, Rails 8.2 changed the #columns method to accept both a single table name (same as <= v8.1) and an array of table names (new). When an array, the returned value changes from `[columns]` to `{table_name => [columns]}`. This detects the two variants and sorts each appropriately.

Original error:
```
$ rails db:schema:dump
NoMethodError: undefined method 'name' for an instance of Array (NoMethodError)

      @connection.columns(...).sort_by(&:name)
                              ^^^^^^^^

Tasks: TOP => db:schema:dump
```

Additionally, the automatic alphabetization of columns introduced in v8.1 has been removed again in v8.2 (rails/rails#56565). This updates the test conditional to be strictly 8.1 instead of >= 8.1.

---

Note: When testing against rails main:HEAD today, I'm seeing a flaky test. I think this is unrelated to the changes in this PR and probably something else in Rails 8.2. It passes sometimes, so might be dependent on test run ordering.
```
  1) Error:
ChangeColumnTest#test_timestamps:
[...]
    test/change_column_test.rb:125:in 'ChangeColumnTest#test_timestamps'
```


ankane:
Hi @zarqman, thanks for the PR. Merged a version of this in the commit above.

ankane:
Also, fixed the timestamp test in eb88bcf3c9b63d3433780a9a79170df021619852.

